PR #1, opened 2020-05-18
Add Event::notify_many

withoutboats:
This API allows the event to notify a fixed number of event listeners greater than 1 (but not all) atomically, during a single lock.

I have use for this in ringbahn, which defines a new synchronization primitive called `AccessQueue`. AccessQueue is like a generalization of a Mutex that limits the number of concurrent accesses to n, instead of 1. (aside: because the limit is greater than 1, it doesn't enable mutable access and is entirely safe code). The usage pattern in ringbahn results in many accesses being released at once. Rather than locking the event repeatedly, it would be ideal to be able to lock it once and notify that many listeners at once.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -99,6 +99,11 @@ impl Inner {
     }
 }
 
+enum Notify {
+    Some(usize),
+    All,
+}
+
 /// A synchronization primitive for notifying async tasks and threads.
 ///
 /// Listeners can be registered using [`Event::listen()`]. There are two ways of notifying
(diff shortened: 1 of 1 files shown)


withoutboats:
Since its an internal API, we could also just use `0` as the sentinel for notify_all, but I thought that would be unclear.

stjepang:
What if this method was named simply `notify()`? Then we'd have:

```rust
event.notify_one(); // same as event.notify(1)
event.notify_all(); // same as event.notify(usize::MAX)
event.notify(n);
```

withoutboats:
That sounds fine!

I've also been mulling over the one-in-one-out vs complete-all-release-all fairness issue we discussed privately. The absolute ideal API (and the most flexible) would be to have an API like this:

```rust
struct Notifier<'a> {
    inner: ListGuard<'a>,
}

impl Event {
    fn notifier(&self) -> Notifier<'_>;
}

impl Notifier {
    fn notify_one(&self);
    fn notify_all(&self);
}
```

That way a user can hold the lock for multiple calls to notify but also interleave other operations with notification. However, it'd be a pretty big API addition, and this pattern of use is rather niche. What do you think about this?

stjepang:
That looks good, we could add such an API in a new PR. I think I'd prefer the following interface though, otherwise the fact that notifier holds a lock on `Event` is not obvious to me:

```rust
impl Event {
    fn lock(&self) -> EventLock<'_>;
}
```

We could also go further and remove the lifetime in this lock guard because we can just bump the refcount inside `Arc`. And if the performance of this atomic bump is a concern, we can also move the reference count into the lock as `refcount: usize` to make it non-atomic.

withoutboats:
I renamed `notify_many` to `notify` and also added a check in its body that `NOTIFIED != 0 || n > 1` so that `notify(1)` and `notify_one()` are truly equivalent.

withoutboats:
This is no longer in your master, what's up?

withoutboats:
It looks like you've completely written the inner notify method without supporting this use case. I've been depending on this API, some communication about what the intent here would've been appreciated.

stjepang:
I'm sorry about that! :(

Realized that listeners were notified in LIFO order rather than FIFO and force-pushed a quick fix in v1.0.1. I also wanted to clean up the code a bit and improve `notify()` and `notify_all()` to bring their time complexities from O(n) down to amortized O(1).

Now I've made the changes I wanted, everything is as efficient as it gets, and the code is a little simpler. Published v1.1.0 so you don't need to use the master branch anymore.

Again, apologies - I hoped I'd get around to sneaking those changes in before you'd notice and it interrupted your work. Should've been upfront instead.

withoutboats:
No worries, thanks!

PR #2, opened 2020-05-19
EventLock API

withoutboats:
Built on top of #1, will rebase after that merges.

I didn't make this take ownership of the Event, because I think that would not be wanted. Users should not be holding the lock on the event for unduly long anyway; having the lock take a lifetime discourages storing the lock for longer than needed. If you feel otherwise, I can revise.

All of the synchronization occurs when the lock is taken in `Event::lock`; if there are no listeners to notify at that point, a lock is simply not taken (rather than holding the lock while performing unrelated operations and doing no-op notifications). I think that's the best performance, but maybe its surprising that the lock isn't always taken - I've documented that fact in the method's docs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -284,6 +284,54 @@ impl Event {
         }
     }
 
+    /// Lock the event to notify multiple listeners.
+    ///
+    /// This API is useful if a client wishes to issue multiple notifications interleaved with
+    /// other operations. This can be useful when a client wishes to notify multiple listeners, but
+    /// fairness would encourage them to perform other operations (such as waking other tasks for a
+    /// different reason) at the same time, rather than doing that before notifying all listeners
+    /// or after.
+    ///
+    /// The lock is held until the `EventLock` guard is dropped, blocking any other threads which
+    /// attempt to register additional listeners. For that reason, users must be cautious not to
+    /// hold the lock for a long period of time.
+    ///
+    /// If there are no active listeners when this is called, the lock will not be acquired.
+    /// 
+    /// # Examples
+    ///
+    /// ```
+    /// # fn call_another_function() { }
+    /// use event_listener::Event;
+    ///
+    /// let event = Event::new();
+    ///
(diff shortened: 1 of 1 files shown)


withoutboats:
rebased

stjepang:
Thanks for the PR! Do you have an example use case for this?

If my thinking is right, this is useful when we want to emit a notification before knowing how many listeners to notify? And I guess we don't want to emit notifications *after* figuring out how many notifications to emit because that might cause spurious wakeups?

withoutboats:
In ringbahn, with `Event::notify`, I first wake all the completed task, then notify that many listeners that there is space on the queue for their submissions. This means that all completed tasks get woken before all tasks waiting to submit work. If the completed tasks perform IO, the waiting tasks get starved.

One option is to wake all waiting tasks before all completed tasks. This doesn't require any new API here, just some code to determine how much space there will be on the queue before processing completed tasks.

Another option is to wake one event listener, then one completed task - one in, one out. This sort of pattern, where the user interleaves other operations with notifications, is what this API enables.

stjepang:
> Another option is to wake one event listener, then one completed task - one in, one out. This sort of pattern, where the user interleaves other operations with notifications, is what this API enables.

`notify(n)` does not wake `n` additional tasks - it only makes sure that among all current listeners, `n` of them are notified. That means `notify(1)` quickly twice in a row is same as just a single `notify(1)`.

Are you taking that into account when interleaving other operations with notifications?

withoutboats:
Ah, you're right because of the change to check NOTIFIED if n is 1.

stjepang:
I think it notifications still don't add up because each `notify(1)` starts from the beginning of the linked list. Tere should be a new method, perhaps named `notify_additional(1)`.

withoutboats:
Ah, got it. I'll close this for now. For my use case, anyway, it's probably better to wake all the waiting tasks with a single notify() call rather than interleaving operations like this.

PR #4, opened 2020-07-01
EventListener::listens_to API

withoutboats:
This is an API for returning whether or not an event listener listens to a given event. This is useful if you want a handle that could be stored in multiple event listener queues over the course of its lifetime, and you need to know which one it is. Otherwise, the handle has to track that state additionally.

I'm not sure about the API surface, so tell me how you think it should look. Right now its just a method on EventListener that takes a reference to an event.

There's also a reasonable desire to have a method that takes another EventListener, and returns whether or not these two listeners listen to the same event.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -507,7 +507,7 @@ impl EventListener {
     /// ```
     #[inline]
     pub fn listens_to(&self, event: &Event) -> bool {
-        ptr::eq(&*self.inner, event.inner.load(Ordering::Relaxed))
+        ptr::eq(&*self.inner, event.inner.load(Ordering::Acquire))
     }
 
     fn wait_internal(mut self, deadline: Option<Instant>) -> bool {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,7 @@ use std::future::Future;
 use std::mem::{self, ManuallyDrop};
 use std::ops::{Deref, DerefMut};
 use std::pin::Pin;
-use std::ptr::NonNull;
+use std::ptr::{self, NonNull};
 use std::sync::atomic::{self, AtomicPtr, AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, MutexGuard};
 use std::task::{Context, Poll, Waker};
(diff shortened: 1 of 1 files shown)


stjepang:
I think another possibility would be to have:

```rust
impl Event {
    fn id(&self) -> EventId;
}

impl EventListener {
   fn event_id(&self) -> EventId;
}

#[derive(Clone, Copy, Eq, PartialEq, Hash)]
struct EventId { ... }
```

This is kind of similar to:

```rust
impl Thread {
    fn id(&self) -> ThreadId;
}

impl<T> JoinHandle<T> {
    fn thread(&self) -> &Thread;
}

#[derive(Clone, Copy, Eq, PartialEq, Hash)]
struct EventId { ... }
```

We could make it even more similar to the thread API if we make `Event` cloneable:

```rust
impl Event {
    fn id(&self) -> EventId;
}

impl EventListener {
   fn event(&self) -> &Event;
}

#[derive(Clone, Copy, Eq, PartialEq, Hash)]
struct ThreadId { ... }
```

Would it help if we went one of those two routes and thus enabled all "equality" checks one could desire at once?

src/lib.rs
```diff
@@ -493,6 +493,23 @@ impl EventListener {
...
+    /// use event_listener::Event;
+    ///
+    /// let event = Event::new();
+    /// let listener = event.listen();
+    ///
+    /// assert!(listener.listens_to(&event));
+    /// ```
+    #[inline]
+    pub fn listens_to(&self, event: &Event) -> bool {
+        ptr::eq(&*self.inner, event.inner.load(Ordering::Relaxed))
```
stjepang, line 510: How important is the `Relaxed` ordering? Would it hurt performance to err on the safe side and go with `SeqCst`?

withoutboats: I'm not sure, I'd be happy to make it SeqCst.

withoutboats: I think it needs to be Acquire, to synchronize with setting the pointer when constructing the first listener (with enough inlining, I think this load could be moved before the store in in `Event::listen` by LLVM if its ordering is Relaxed).

withoutboats:
@stjepang The only problem with `EventId` is that Event would probably need to return `Option<EventId>` (and `EventId` would be a `NonZeroUsize`), because we don't want to force an allocation to get an eventid from an event, but we don't want different events to have the same id. This might be more trouble than its worth.

stjepang:
Maybe `EventId` is too much :) Okay, I feel good about `EventListener::listens_to()`.

*thinking out loud* The counterpart that checks whether two `EventListener`s listen to the same event could be named `EventListener::same_event()` (`futures-channel` has similarly named [`Sender::same_receiver()`](https://docs.rs/futures-channel/0.3.5/futures_channel/mpsc/struct.Sender.html#method.same_receiver)).

stjepang:
Published v2.1.0 containing this PR.

Issue #7, opened 2020-07-24
poll not waking when Pending is returned


Licenser:
This is a follow up to https://github.com/async-rs/async-std/issues/841 I tracked the poll that isn't waking in the chain down to the EventListener.

It seems that when pending is returned 

```
cx.waker().wake_by_ref();
```

needs to be called or having more event listeners active then executer threads seems to lead to them indefinitively spinning.

I'll add a PR for this.

stjepang:
What's the status here? Thoughts on this comment? https://github.com/async-rs/async-std/issues/834#issuecomment-663507266

My understanding is that this is not a bug, but I could be misunderstanding.

Licenser:
The two things are independent of each other. The problem here, from what I understood, is that if there are enough event-listeners they could block the entire pool since they'll never wake another task when they're pending.

stjepang:
Hmm I'm a bit confused by that PR. An `EventListener` is notified from `notify()` by waking the waker stored in its state.

If an `EventListener` is polled and hasn't been notified yet, we need to return `Pending`. I don't understand why we would wake the waker at that point. Won't that reschedule the same task that will likely just poll `EventListener` again and keep doing that forever?

Licenser:
Hmm, that makes sense, I understood the docs for the weaker differently - I thought it would need to be called to **not** reschedule the same task :); yes then this makes no sense at all.

PR #8, opened 2020-07-24
Wake on Pending

Licenser:
This fixes the issue mentioned in #7


stjepang:
I think this can be closed now too :)

Thank you for investigating this problem, I appreciate it very much!

Licenser:
Yes, sorry, I forgot to close the PR :) and many thanks for the patience of explaining my why this was wrong :D <3

PR #13, opened 2020-08-25
Extend Events to support tokens for listeners.

withoutboats:
The `Event` and `EventListener` types are parameterized by a new type `T`, which represents the type of the token that is passed with each notifitication. This type is `()` by default, unless specified otherwise by users. This allows users to pass tokens from the notifier to the listener with new `notify_with_token` APIs. The `EventListener` Future now resolves to its token type, instead of always resolving to `()`.

An example use case is a `Mutex` which, like the parking_lot Mutex, can pass the lock directly on to the next waiting listener, without ever actually unlocking the Mutex's state.

This change was done backwards compatibly, so APIs which previous relied on the absence of a token type (like the existing notify APIs, and some of the wait APIs) are only provided by events which have no token, whereas new APIs with tokens are added additionally.

Events with tokens can only notify one event at a type. It would be interesting to extend this to support passing multiple tokens to notify multiple listeners at once, with tokens. The best API I can think of would rely on `min_const_generics` (and so be unsuitable as an additional right now) but would look like this:

```rust
impl<T> Event<T> {
     pub fn notify_with_tokens<const N: usize>(&self, tokens: [T; N]) {
          // ...
     }
}
```

@stjepang This is just an idea I had that seemed like an easy extension. I don't have any compelling and urgent need for this functionality (yet, at least!) If you're not interested in supporting it, that's cool with me, I just wanted to explore the possibility.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -78,19 +78,19 @@ use std::time::{Duration, Instant};
 use std::usize;
 
 /// Inner state of [`Event`].
-struct Inner {
+struct Inner<T> {
     /// The number of notified entries, or `usize::MAX` if all of them have been notified.
     ///
     /// If there are no entries, this value is set to `usize::MAX`.
     notified: AtomicUsize,
 
     /// A linked list holding registered listeners.
-    list: Spinlock<List>,
+    list: Spinlock<List<T>>,
 }
 
-impl Inner {
+impl<T> Inner<T> {
     /// Locks the list.
-    fn lock(&self) -> ListGuard<'_> {
+    fn lock(&self) -> ListGuard<'_, T> {
         ListGuard {
             inner: self,
             guard: self.list.lock(),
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -117,22 +117,26 @@ impl Inner {
...
     /// A pointer to heap-allocated inner state.
     ///
     /// This pointer is initially null and gets lazily initialized on first use. Semantically, it
     /// is an `Arc<Inner>` so it's important to keep in mind that it contributes to the [`Arc`]'s
     /// reference count.
-    inner: AtomicPtr<Inner>,
+    inner: AtomicPtr<Inner<T>>,
 }
 
+// TODO
```
withoutboats, line 129: Haven't bothered figuring out yet what the bounds need to be on auto traits.

src/lib.rs
```diff
@@ -701,72 +775,75 @@ impl Drop for ListGuard<'_> {
...
-    Notified(bool),
+    Notified(bool, T),
 
     /// An async task is polling it.
     Polling(Waker),
 
     /// A thread is blocked on it.
     Waiting(Thread),
+
+    Dummy,
```
withoutboats, line 810: I added this `Dummy` state since `Notified` can't be used as a dummy state. `Created` could instead, but I thought having the panics is better security in depth. Let me know if you avoided this intentionally because the branches hurt benchmarks or something like that.

src/lib.rs
```diff
@@ -862,6 +939,46 @@ impl List {
...
+
+        self.notify_internal(false, token);
+    }
+
+    fn notify_additional_with_token(&mut self, token: T) {
+        self.notify_internal(true, token);
+    }
+
+    #[inline(always)]
+    fn notify_internal(&mut self, additional: bool, token: T) -> bool {
```
withoutboats, line 955: I separated the logic for notification on `List` out into this single function which all the notification functions on List call into.

src/lib.rs
```diff
@@ -870,58 +987,16 @@ impl List {
         }
         n -= self.notified;
 
-        while n > 0 {
+        while n > 0 && self.notify_internal(false, ()) {
```
withoutboats, line 990: These while loops are probably a bit too cute. Happy to rewrite them as this if you think mutating state in the conditional is too confusing:

```rust
while n > 0 {
    n -= 1;
    if self.notify_internal(false, ()) {
        break;
    }
}
```

stjepang:
Sorry I haven't had the time to review this yet, been kinda swamped for the last two weeks :(

Glancing over it, this is a change I've also wanted before so it seems good to me. I also feel open to adding new APIs and experimenting with this crate in general.

Would you perhaps like push/publish permissions on this crate so you're not always blocked on me?

withoutboats:
@stjepang I wouldn't mind, but I haven't been blocked on this at all. It was also just an idea I wanted to explore!

notgull:
Superseded by #52

PR #15, opened 2020-09-21
EventListener: Add pub fn `discard()`

kalcutter:
This function provides a convenient way to drop a listener and avoid
having its notification sent to another active listener.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -522,6 +522,36 @@ impl EventListener {
         self.wait_internal(Some(deadline))
     }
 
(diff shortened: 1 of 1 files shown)


kalcutter:
I just noticed the token PR. If you are serious about adding support for token events, then the API should be tweaked a little to fit into that model. In that case, maybe a signature like `pub fn take() -> Option<T>` would be more appropriate.

stjepang (approved):
This seems good to me. Thanks!

stjepang:
Published v2.5.0

PR #27, opened 2022-08-23
Use the parking crate instead of threading APIs

notgull:
In order to avoid the downsides of user-invoked `thread::Thread::unpark()`, this PR imports the `Parking` crate for use in parking/unparking threads.

Is there a reason why the original author elected to use `std::thread` instead of `parking`?


notgull:
Ah, that's why, the MSRV bump. Although it's only a bump of 3 versions, up to a version in the 1.30's. Are we fine with that bump?

taiki-e (approved):
LGTM

PR #29, opened 2022-08-29
Bump rust-version to 1.39

notgull:
Forgot to do this after #27


taiki-e:
Could you do the same to `.clippy.toml`?

https://github.com/smol-rs/event-listener/blob/a01518f41c5c4acd71fb6a55ef2cc2675310dfca/.github/workflows/ci.yml#L37-L39

taiki-e (approved):

PR #30, opened 2022-09-05
Implement a new algorithm to make this crate `no_std`

### LINKED ISSUE #28: Make this crate `no_std`
If this crate were `no_std`, it would allow crates like `async-lock` and `async-channel` to be used on `no_std` as well. I envision an API with a default `std` feature, and without it this crate would not depend on `parking` (and the `wait` API would disappear as well).

However, the current version of the code uses a `Mutex` to lock the linked list containing event listeners. Theoretically, we could replace this with an atomic linked list similar to the one used in `concurrent-queue`.

This would help `smol` be able to be used in embedded systems. Any thoughts on this?

notgull:
This PR resolves #28 by replacing this crate's algorithm with one that is compatible with `no_std`. I realized that the current algorithm is basically a concurrent queue, so I replaced it with our `concurrent_queue` crate. The underlying implementation is basically the same, but there is a new `std` feature that prevents `parking` primitives from being used. This is a breaking change, since removing the `std` feature removes the `wait_*` family of functions.

For now, this is a draft PR because:

- I'm waiting for the changes I made to `concurrent_queue` to be released; right now I'm just using ~~master~~ the branch where I'm implementing Loom support.
- ~~The new implementation pasts the existing tests, but I'd like to check to see if it also passes the tests for `async-channel` and `async-lock` when it is patched in.~~ Test suites pass when I patch in this new version of `event-listener`.
- ~~There's a merge conflict I don't feel like resolving right now.~~ Should be fixed by now.
- ~~I'd like to benchmark the changes before I merge them. The new algorithm may allocate more often than the older algorithm. Although this allocation will likely be amortized over the program's runtime, I'd like to be sure that it doesn't impact performance significantly. See #31.~~ See my comment below.
- I also implemented partial support for `loom` testing, but I haven't gotten around to writing the tests for that yet.

Future considerations:

- Since `ConcurrentQueue::bounded()` is more efficient than `ConcurrentQueue::unbounded()`, it may be useful to provide a "with maximum length" method that allocates everything up front.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -392,6 +392,19 @@ impl fmt::Debug for Event {
     }
 }
 
+impl Drop for Event {
+    fn drop(&mut self) {
+        // Drop the inner state.
+        self.inner.with_mut(|inner| {
+            if !inner.is_null() {
+                unsafe {
+                    drop(Arc::<Inner>::from_raw(*inner));
+                }
+            }
+        });
+    }
+}
+
 /// A guard waiting for a notification from an [`Event`].
 pub struct EventListener {
     /// The reference to the original linked list.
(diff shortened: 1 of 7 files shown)


notgull:
Now that I can test the benchmarks, it seems that it's regressed to an almost unacceptable degree.

benching

I'm not entirely sure where the extra overhead is coming from, I'll start looking into it.

taiki-e:
Seems Miri failure is due to detached threads in doctest: https://github.com/rust-lang/miri/issues/1371

A workaround that I sometimes use is adding sleep to the end of test: https://github.com/crossbeam-rs/crossbeam/pull/891

notgull:
With the newest commit, I have it down to only 30% slower.

src/sync.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub(crate) use loom::thread;
+
+    pub(crate) mod atomic {
+        pub(crate) use core::sync::atomic::compiler_fence;
+        pub(crate) use loom::sync::atomic::*;
+    }
+
+    use loom::sync::{Arc, Condvar, Mutex};
+
+    /// Re-implementation of `parking::pair` based on loom.
```
zseri, line 63: Can we upstream this into https://github.com/smol-rs/parking ?

notgull: Sure! It should be a straightforward reimplementation in terms of Loom primitives.

src/lib.rs
```diff
@@ -418,242 +408,215 @@ impl Event {
...
-                        // Remove the entry and check if notified.
-                        return self
-                            .inner
-                            .lock()
-                            .remove(entry, self.inner.cache_ptr())
-                            .is_notified();
+                    #[cfg(loom)]
+                    {
+                        let _ = deadline;
+                        panic!("`wait_deadline` is not supported with loom");
```
zseri, line 603: why do we not make this a compile-time error?

notgull: This function is still used by the `wait()` function, even though the timeout in that case will always be `None`. A compiler error would prevent that from happening.

zseri: Can we refactor it in such a way that `timeout` optionality is implemented via traits or such, instead of an `Option`, oh, and how does that interact with `no_std`, or do `wait` functions just not work there?

zseri: I know this would "blow up" the API even more, but would imo be preferable.

notgull: I could implement `timeout` as a trait object that selects between a `NoTimeout` option and a `Timeout(Duration)` option, but that feels like something that's out of scope for this PR.

Yes, there is no way to port the `wait()` functions to `no_std`, unless we wanted to try to port `parking` to `no_std`. IIRC that would be impossible unless we use [`critical-section`](https://crates.io/crates/critical-section) or something like that, and in my opinion, we shouldn't.

src/sync.rs
zseri: now that it's upstreamed, can we use that version? (although I don't know how it would interact with dependencies...) Does cargo correctly work with git deps which also have versions (e.g. transforms them into crates.io entries on publishing)?

notgull:
Closing in favor of #33

PR #33, opened 2022-11-01
Implement an algorithm to make this crate no_std, take 2

notgull:
I'm not satisfied with the direction #30 has taken, so I've written this PR. It's less destructive, less verbose, and faster than the previous one. I haven't implemented `portable_atomic` or `loom` support here, I'll do that in future PRs.

In this PR, rather than using a `Mutex` to wrap an unsync linked list (what's currently on `master`) or using `ConcurrentQueue` (as in #30), I use an atomic singly linked list to store listeners. Similarly to #30, listeners in this list can be marked as "orphaned" to indicate that they should not be notified (e.g. when the `EventListener` is dropped). As opposed to #30, it only exhibits a 20% slowdown as compared to `master` on my PC. With the increased number of atomic operations involved, I think that that is about as low as we can go without using a locking mechanism of some kind.

Future work:

- Add `portable_atomic` and `loom` support.
- It may be possible to write this crate using only safe code. I intentionally separated out `RacyArc` because we may be able to split it off into another crate, the interior mutability in `Listener` can be resolved using `crossbeam_utils::AtomicCell<Wakeup>` or similar, and the linked list could probably be emulated using `AtomicCell` and `Arc` as well. The main concern is that we would have to not use the cache in `ListenerQueue`, and this (among other things) would likely cause significant performance issues (`Arc` is expensive).
- Another way of doing this is using a spinlock, but making sure we never actually spin on it. We keep the `master` branch, but replace the `Mutex` with a spinlock. `notify` and `listen` both try to access the spinlock. If `notify` sees the spinlock active, it places the notification counts in an atomic variable and whoever holds the spinlock preforms the notification once they're about the drop the lock. If `listen` (or any of `EventListener`'s methods) see the spinlock locked, they put their `Waker`s (or `Unparker`s) into a short-term `ConcurrentQueue` which is then used to set up listeners once the current holder tries to drop the lock. This may be a better way of doing it, but I don't know whether it would be faster or slower.

### DIFF
--- a/src/listener.rs
+++ b/src/listener.rs
@@ -1,0 +1,410 @@
+//! A listener in the queue of listeners.
+
+use crate::sync::atomic::{AtomicUsize, Ordering};
+use crate::sync::cell::Cell;
+
+use core::task::Waker;
+
+#[cfg(feature = "std")]
+use parking::Unparker;
+
+/// A listener in the queue of listeners.
+pub(crate) struct Listener {
+    /// The current state of this listener.
+    ///
+    /// This also serves as the refcount:
+    /// - Ref 1: The listener state is not orphaned.
+    /// - Ref 2: The listener is in the queue (queue bit is set).
+    state: AtomicUsize,
+
+    /// The waker or thread handle that will be notified when the listener is
+    /// ready.
+    ///
+    /// This is kept synchronized by the `state` variable; therefore, it's
+    /// technically `Sync`.
+    waker: Cell<Wakeup>,
+}
+
+impl Listener {
+    /// Create a new listener.
+    pub(crate) fn new() -> Self {
+        Self {
+            state: AtomicUsize::new(
+                State {
+                    listen: ListenState::Created,
+                    queued: false,
+                }
+                .into(),
+            ),
+            waker: Cell::new(Wakeup::None),
+        }
+    }
+
+    /// Enqueue this listener.
+    pub(crate) fn enqueue(&self) {
+        let mut state = State::from(self.state.load(Ordering::Acquire));
+
+        // If the listener is already queued, then we don't need to do anything.
+        loop {
+            if state.queued {
+                return;
+            }
+
+            // Mark the listener as queued.
+            let new_state = State {
+                queued: true,
+                ..state
+            };
+
+            match self.state.compare_exchange(
+                state.into(),
+                new_state.into(),
+                Ordering::AcqRel,
+                Ordering::Acquire,
+            ) {
+                Ok(_) => break,
+                Err(actual) => state = State::from(actual),
+            }
+        }
+    }
+
+    /// Dequeue this listener.
+    ///
+    /// Returns `true` if the listener is also orphaned, and that the caller
+    /// should drop the listener.
+    pub(crate) fn dequeue(&self) -> bool {
+        let mut state = State::from(self.state.load(Ordering::Acquire));
+
+        // If the listener is not queued, then we don't need to do anything.
+        loop {
+            if !state.queued {
+                return false;
+            }
+
+            // Mark the listener as not queued.
+            let new_state = State {
+                queued: false,
+                ..state
+            };
+
+            match self.state.compare_exchange(
+                state.into(),
+                new_state.into(),
+                Ordering::AcqRel,
+                Ordering::Acquire,
+            ) {
+                Ok(_) => break,
+                Err(actual) => state = State::from(actual),
+            }
+        }
+
+        // If the listener is orphaned, then we need to drop it.
+        state.listen == ListenState::Orphaned
+    }
+
+    /// Orphan this listener, and return `true` if we need to be dropped.
+    ///
+    /// This method is called when `EventListener` is dropped. The second result is whether or not
+    /// the listener was notified (and `true` if the notification was additional.)
+    pub(crate) fn orphan(&self) -> (bool, Option<bool>) {
+        let mut state = State::from(self.state.load(Ordering::Acquire));
+
+        loop {
+            match state.listen {
+                ListenState::Orphaned => {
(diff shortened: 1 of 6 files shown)


src/listener.rs
```diff
@@ -0,0 +1,410 @@
...
+    /// - Ref 1: The listener state is not orphaned.
+    /// - Ref 2: The listener is in the queue (queue bit is set).
+    state: AtomicUsize,
+
+    /// The waker or thread handle that will be notified when the listener is
+    /// ready.
+    ///
+    /// This is kept synchronized by the `state` variable; therefore, it's
+    /// technically `Sync`.
+    waker: Cell<Wakeup>,
```
zseri, line 25: wouldn't it make more sense to use `UnsafeCell` here?

src/queue.rs
```diff
@@ -0,0 +1,224 @@
...
+        // Is this node the current cache node?
+        if ptr::eq(self.cache.node.get() as *const Node, node.as_ptr()) {
+            // We can now clear the cache.
+            unsafe { (self.cache.node.get() as *mut Node).drop_in_place() };
+            self.cache.occupied.store(false, Ordering::Release);
+        } else {
+            // We need to deallocate the node on the heap.
+            unsafe { Box::from_raw(node.as_ptr()) };
+        }
+    }
```
zseri, line 203: these methods (`alloc`, `dealloc`, `orphan`) appear to only act on the `CachedNode` and imo should be moved into an `impl` block associated with that. (and if these methods are still necessary, should be replaced with forwarding methods)

src/util.rs
```diff
@@ -0,0 +1,83 @@
...
+
+use crate::sync::atomic::{AtomicPtr, Ordering};
+use crate::sync::{Arc, AtomicWithMut};
+
+use core::mem::ManuallyDrop;
+use core::ptr;
+
+/// A wrapper around an `Arc` that is initialized atomically in a racy manner.
+#[derive(Default)]
+pub(crate) struct RacyArc<T> {
```
zseri, line 11: How big is the difference between this and `OnceCell` impls?

notgull: Two main differences:

1). While the once cells are synchronized, this one is racy. While the once cells only run one closure at a time, this one runs all closures at once and sets the value to whichever one comes first.

2). It's only one system word in length, while once cells are usually at least three.

In addition, since our once cells are backed by this crate, we can't use once cells here.

notgull: We're trying to avoid the `once_cell` crate due to its recent MSRV bump, see smol-rs/async-io#93

src/listener.rs
notgull: I originally used `UnsafeCell`, but given that most of the semantics involved are just "replace the `waker` with something else", I ended up using `Cell` instead.

I tried replacing it with `UnsafeCell<MaybeUninit<Wakeup>>` (and also getting rid of the `Wakeup::None` variant), but any gains are minimal at best. I figure it's better to just use `Cell` and avoid the unsafe code.

src/queue.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn alloc(&self, listener: Listener) -> NonNull<Node> {
+        // Try to install a cached node.
+        if !self.cache.occupied.swap(true, Ordering::Acquire) {
+            // We can now initialize the node.
+            let node_ptr = self.cache.node.get() as *mut Node;
+
+            unsafe {
+                // Initialize the node.
+                node_ptr.write(Node::new(listener));
+
```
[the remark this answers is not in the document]
notgull: I bypass it by just casting `node.get()` into an `*mut Node`, which allows me to write the `Node` straight into the slot without worrying about `MaybeUninit`.

src/queue.rs
notgull: I've moved `alloc` and `dealloc`, but `orphan` is, semantically, an important operation on the `Queue` so I've kept it there.

src/queue.rs
```diff
@@ -0,0 +1,226 @@
...
+            // We can now initialize the node.
+            let node_ptr = self.node.get() as *mut Node;
+
+            unsafe {
+                // Initialize the node.
+                node_ptr.write(Node::new(listener));
+
+                // Return the node.
+                NonNull::new_unchecked(node_ptr)
+            }
```
zseri, line 186: I don't like this part at all, especially since it appears to misuse `MaybeUninit` and/or `UnsafeCell` somewhat. how about:
```rust
unsafe {
    let node_ptr: &mut MaybeUninit<_> = &mut *self.node.get();
    // Initialize the node
    node_ptr.write(Node::new(listener));
    // Return the node
    node_ptr.assume_init_mut().into()
}
```

src/util.rs
zseri: I'd like to ask about the performance of this, because in the past I implemented https://github.com/zseri/revenq,which I originally based upon a similar pattern and hit a huge perf bottleneck. Thus I ask: How much does this get "hammered" during its lifetime or the lifetime of the chain?

notgull: After initialization (which is a one-time process), loads from the `AtomicPtr` are a simple atomic load. This is opposed to a `OnceCell`, which consists of a load as well. I doubt that there's much overhead aside from at initialization time. At worst, this leads to the initialization code being run once per thread.

I could replace it with the proper cell to see what happens, but I doubt it would make a difference. Besides, this is what this code used prior to this change.

src/queue.rs
notgull: I've replaced the node cast to an `as_mut_ptr()` usage. Unfortunately I can't use `MaybeUninit::write()` and `MaybeUninit::assume_init_mut()` since they we introduced in Rust 1.55, which would lead to a significant MSRV bump

notgull:
Closing in favor of #34

PR #34, opened 2022-11-05
Implement an algorithm to make this crate no_std, take 3

notgull:
This PR supersedes #33 and aims to improve on its strategy using a hybrid between its strategy and the strategy that's currently used. Basically, it keeps the `master` strategy, but replaces the `Mutex` with another implementation based on the event queue in #33.

To clarify, the current strategy used on the `master` branch is to use the `std::sync::Mutex` type to wrap a `!Sync` linked list built using the `Cell` type. At its core, `event-listener` is just a queue of wakers that we wake up whenever we `notify` the event. On modern operating systems, this is very fast. Mutexes are implemented using futexes, meaning that in the case of no contention there are, at most, two atomic operations involved. One to lock the mutex, and one to unlock. The main disadvantage of this approach is that the `Mutex` requires `libstd` to be linked, which forces the library (and anything that depends on it) to not be `no_std`. One of my goals is to be able to use `smol` on embedded systems without needed to resort to [`embassy`](https://github.com/embassy-rs/embassy), an RTOS or something similar. While `async-io`, `blocking` and their dependents are not usable on these platforms, it would be nice to at least have an `async` toolkit for mutexes, et al.

My strategy for #30 and #33 was to reimplement `event-listener` by replacing this list with a fully atomic list. In #30, I used the `concurrent-queue` crate to implement the listener queue, and in #33 I used a homegrown singly atomic linked list implementation. The main advantage of these implementations is that they are completely lock-free. Although I didn't test for this, I imagine that in both of these implementations, worst-case contention would be reduced significantly. In addition, since they only used atomics and not locks, they were available on `no_std` platforms. In fact, by taking advantage of [`portable-atomic`](https://crates.io/crates/portable-atomic), it could even run on platforms without atomic support.

The primary disadvantage of these approaches was speed. #30 exhibited a slowdown of up to 50%, which was completely unacceptable. Even through certain optimizations, I couldn't get it to break 30%. #33 did better; at best, it only exhibited a 15% slowdown. Still, it's imposing a 15% performance tax on all users, including the admittedly larger pool of use cases that would only ever use `libstd`. The ideal solution would be to not incur any performance issues over the current system.

A theoretically ideal solution would be to replace the `Mutex` with a spinlock. It's unlikely to occur much contention anyways, so in the ideal case (and, arguably, the overwhelming majority of usecases) it would be identical to the current system while also being `no_std`. However, spinlocks are [dangerous](https://matklad.github.io/2020/01/02/spinlocks-considered-harmful.html), and a [non-starter](https://github.com/smol-rs/event-listener/issues/28#issuecomment-1229713778) anyways.

This solution began with the idea that we could use this theoretical spinlock in the best case and then fall back to the listener queue case when there's too much contention. This "hybrid" approach would allow us to utilize the performance gains of the spinlock while also being able to fall back onto the queue for the worst cases. While this approach is difficult to directly implement in Rust, I've created a modified version of this approach that should be able to emulate it well enough.

This new implementation is based on the idea that, in the ideal use case, the inner `Mutex` should never be contended. The `Mutex` in the `Event` structure is replaced with an optimistic spinlock. In the best case, the crate functions much like it does today It will try to lock the mutex, and if it does, performs the desired operation (e.g. inserting into the listener list, notifying an entry, etc.). However, if the mutex is locked, it pushes a node representing the desired operation onto the backup queue. The operations that can be performed are as follows:

- Push a listener into the list.
- Notify listeners.
- Remove a listener from the list, and optionally propagate its notification if it has one.
- Wake a waker. There are some operations that I cannot effectively emulate with this system, so the waker just waits for the mutex to be available.

Once whoever holds the mutex lock is about to unlock it, they check this backup queue. If there are any operations, they are popped off of the queue and ran.

There is definitely room for improvement in this system, but speed-wise, I've accomplished my goal. From benchmarks, it's at least on par with the `master` implementation, and even seems to be faster in certain cases.

Future work would involve making the spinlock fair, as well as implementing `portable_atomic` and `loom` features.

~~Things I need to do before marking this PR as ready to review:~~

- ~~Determine semantics.~~
- ~~The new event queue needs more testing.~~
- ~~Clean up code overall.~~
- ~~MIRI is failing, determine why.~~ ~~MIRI seems to be deadlocking? I'm not sure why, maybe it's a MIRI bug?~~ With the new queue implementation this issue appears to have vanished.


notgull:
MIRI is deadlocking for some reason? I can't tell if it's the code or if it's a MIRI bug. Anyways, I've opened this for review now.

zseri:
Cargo hack build does not seem to support `--no-dev-dep` (anymore?) idk what's up with that, but it makes CI fail.

notgull:
I just took a look, I typo'd `no-dev-deps` (with an `s` at the end), let me fix that.

src/inner.rs
```diff
@@ -0,0 +1,224 @@
...
+                return Some(MutexGuard { mutex: self });
+            }
+
+            // Use atomic loads instead of compare-exchange.
+            while self.locked.load(Ordering::Relaxed) {
+                if spins <= 0 {
+                    return None;
+                }
+
+                spins -= 1;
```
zseri, line 196: I think it would be nice if this would be replaced with `.checked_sub` + make `spins: u32`.

src/queue.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::node::Node;
+use crate::sync::atomic::{AtomicPtr, Ordering};
+
+use crossbeam_utils::CachePadded;
+
+use alloc::boxed::Box;
+use core::ptr;
+
+/// A queue of nodes.
+pub(crate) struct Queue {
```
zseri, line 12: why can't we use `concurrent-queue` for this?

src/lib.rs
```diff
@@ -1001,3 +874,42 @@ fn full_fence() {
...
+    #[cfg(not(feature = "std"))]
+    #[allow(deprecated)]
+    sync::atomic::spin_loop_hint();
+}
+
+impl Event {
+    /// Locks the event.
+    ///
+    /// This is useful for simulating contention, but otherwise serves no other purpose for users.
+    /// It is used only in testing.
```
zseri, line 893: "it is used only in testing" then it should be behind a feature flag.
e.g. `#[cfg(any(feature = "test", test))]`

src/node.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /// Create a new waiting entry.
+    pub(crate) fn waiting(task: Task) -> Self {
+        NodeData::Waiting(task).into()
+    }
+
+    pub(crate) fn next(&self) -> &AtomicPtr<Node> {
+        &self.next
+    }
```
zseri, line 82: make `.next` `pub(crate)` instead, the pointer is atomic, so that we only allow `&` access here doesn't seem to matter.

src/node.rs
```diff
@@ -0,0 +1,147 @@
...
+                // Remove the listener from the list.
+                let state = list.remove(listener, inner.cache_ptr());
+
+                if let (true, State::Notified(additional)) = (propagate, state) {
+                    // Propagate the notification to the next listener.
+                    if additional {
+                        list.notify_additional(1);
+                    } else {
+                        list.notify(1);
+                    }
```
zseri, line 127: here we have basically code to decide between `notify_additional` vs `notify`. This should be factored out.

src/inner.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub(crate) notified: AtomicUsize,
+
+    /// A linked list holding registered listeners.
+    list: Mutex<List>,
+
+    /// Queue of nodes waiting to be processed.
+    queue: Queue,
+
+    /// A single cached list entry to avoid allocations on the fast path of the insertion.
+    cache: UnsafeCell<Entry>,
```
zseri, line 30: document how access to this field is locked when necessary. (that is, what is the invariant under which that field can be modified and/or read)

zseri (changes requested):

src/queue.rs
```diff
@@ -0,0 +1,102 @@
+//! The queue of nodes that keeps track of pending operations.
+
+use crate::node::Node;
+use crate::sync::atomic::{AtomicPtr, Ordering};
+
+use crossbeam_utils::CachePadded;
```
zseri, line 6: `concurrent-queue` explicitly uses `cache-padded` for this instead of `crossbeam-utils`, this crate should do the same as it is usually used together with `concurrent-queue`

src/queue.rs
notgull: I used `concurrent-queue` in #30 and it led to a significant slowdown compared to the current implementation. Although, the new queue isn't on the hot path anymore so I doubt it matters that much.

Also, `concurrent-queue` hasn't released a version yet with the `no_std` patch. I can file a PR for that if you want.

src/queue.rs
notgull: On the current `master` branch of `concurrent-queue`, it's been ported to use `crossbeam-utils`.

https://github.com/smol-rs/concurrent-queue/blob/master/src/unbounded.rs

src/lib.rs
notgull: Adding this fails with:

```
error[E0599]: no method named `__lock_event` found for struct `Event` in the current scope
  --> tests/queue.rs:24:22
   |
24 |     let lock = event.__lock_event();
   |                      ^^^^^^^^^^^^ method not found in `Event`

For more information about this error, try `rustc --explain E0599`.
```

I've added a `__test` feature that should circumvent this issue that is enabled in testing.

src/queue.rs
zseri: file a PR please, I would really want to avoid code duplication/diverging optimizations between multiple smoi-rs crates.

src/queue.rs
zseri: ah ok.

src/queue.rs
notgull: Here it is: smol-rs/concurrent-queue#29

zseri: What is the difference between the algorithms used in this queue impl vs. the one in `concurrent-queue`? That is, are here any improvements contained which could be upstreamed into `concurrent-queue`?

notgull: The one used in this queue impl is a fairly naive algorithm that allocates per item. The main advantage of this algorithm over something like `concurrent-queue` is the fact that, in the best case scenario, there are fewer atomic operations. I would not recommend upstreaming this `concurrent-queue`.

notgull: With version 2.0.0 of `concurrent-queue`, I've updated this PR to use `concurrent-queue`.

zseri:
```
error[E0599]: no method named `__lock_event` found for struct `Event` in the current scope
  --> tests/queue.rs:24:22
   |
24 |     let lock = event.__lock_event();
   |                      ^^^^^^^^^^^^ method not found in `Event`
```

zseri:
Suggestion: https://github.com/smol-rs/event-listener/commit/9c31ae670a9755af5965b61ee673e891f99e31df (refactors `Node` such that the `Drop` handler is moved to another struct, such that the `Node` itself can be freely deconstructed, and moves the dealing with the `Option` to the place where it actually matters)

notgull:
> Suggestion: [9c31ae6](https://github.com/smol-rs/event-listener/commit/9c31ae670a9755af5965b61ee673e891f99e31df) (refactors `Node` such that the `Drop` handler is moved to another struct, such that the `Node` itself can be freely deconstructed, and moves the dealing with the `Option` to the place where it actually matters)

Great idea! I've merged your commit into this branch.

src/lib.rs
```diff
@@ -60,52 +60,83 @@
 //! }
 //! ```
 
+#![no_std]
```
zseri, line 63: wasn't it problematic in other crates when this was always set even if `feature="std"` is set?

zseri (changes requested):
Maybe in modules which aren't `pub`-reachable from the top module, we should replace `pub(crate)` with `pub` (which are equivalent in that case), reducing clutter.

src/list.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+
+    /// Get the state of this entry.
+    pub(crate) fn state(&self) -> &Cell<State> {
+        &self.state
+    }
+
+    /// Tell whether this entry is currently queued.
+    #[cfg(feature = "std")]
```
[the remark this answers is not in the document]
notgull: It is only ever used as an optimization for the `park_internal` function, I'll indicate as such.

notgull:
> Maybe in modules which aren't `pub`-reachable from the top module, we should replace `pub(crate)` with `pub` (which are equivalent in that case), reducing clutter.

I generally like to make it `pub(crate)` even if it's not reachable from the outside, for two reasons:

* Makes it slightly harder to have API leaks.
* I usually program with `warn(clippy::pedantic)`, which issues a warning if it sees a `pub` where a `pub(crate)` would work.

notgull:
I also fixed the MIRI deadlock... does MIRI not support `compare_exchange_weak`? Weird.

zseri:
As the problem with the previous iterations was a bigger regression, and the current benchmarks indicate that that is no longer the case (the new branch and old branch have comparable iteration time, and the noise is relatively strong in both). Can we squash-merge this?

zseri:
Something I'd like to do next is bumping `criterion`, as this massively reduces the amount of pulled-in crates on dev-builds.

notgull:
> Can we squash-merge this?

Sounds good!

zseri (approved):

PR #35, opened 2022-11-11
Bump criterion to 0.4.0

notgull:


zseri:
Note: https://github.com/smol-rs/event-listener/commit/87fd707b988f6f1e445fc27a8b765f38e77c1673 it would be nice to add
```toml
default-features = false
features = ["cargo_bench_support"]
```
which should make the set of pulled in deps smaller (at least in the future)... I'm not sure which additional features might also be necessary.. see also https://github.com/bheisler/criterion.rs/blob/326ebbb2fe48fbb683096fb1cf680a759b35a27e/Cargo.toml

zseri:
`cargo-criterion` will later be preferred for generating reports, moving that part out of the stuff which gets compiled for the/each crate tested by criterion. `cargo_bench_support` well, is necessary to prevent future breaks of `cargo bench` support.

notgull:
Hmm, I'd probably prefer merging the `criterion-bump` branch over this one.

zseri (approved):

zseri:
ok, next up is probably adjusting `.gitignore` and `Cargo.toml` so that they're flamegraph-friendly.

PR #37, opened 2022-11-11
Refactor RacyArc out of Event

notgull:
The `Event` contains code for lazily initializing the `Arc<Inner>` structure using an atomic, racy setup condition. I feel like this could be refactored into another module, since it's separate from the `Event`'s business logic. Really, this could probably be separated into another crate entirely, as other `smol` crates (e.g. `async-executor`) could probably make use of it as well. Until then, however, I've moved it into another module.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,6 +73,7 @@ mod list;
 mod node;
 mod queue;
 mod sync;
+mod util;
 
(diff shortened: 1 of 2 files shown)


notgull:
Coming back to this, I think I'll pass on this.

PR #38, opened 2022-11-11
Replace the linked list with a safer and less allocation-heavy alternative

notgull:
Right now, the current algorithm on `master` uses a linked list to implement the queue of listeners that powers the `Event` structure. The allocation scheme for this linked list is relatively naive and unsafe. In the fast case, it writes to an `UnsafeCell` cache entry inside the `Event`'s allocation proper. In the slow case, it allocates (and then deallocates) a `Box` heap allocation containing the `Entry`.

This PR replaces it with an algorithm that is both safer and more efficient. It adds the `slab` crate as a dependency and uses that to store the `Entry`s. In addition, to keep track of the `Entry`s, this crate no longer stores `NonNull<Entry>`s, but instead the index of the `Entry` in the `Slab`. This strategy allows us to eliminate a significant amount of unsafe code.

I've also modified the `Node` code a bit to accommodate this new change. This led to the introduction of a `TaskWaiting` structure that the `Node` uses to communicate with the `EventListener`. This also reduces the amount of unsafe code involved.

Benchmark wise, it's about equal to the current algorithm, plus or minus some level of noise.

noise

Here are quantitative measures of the unsafe code eliminated:

Current `master`:

master

New branch:

new branch

### DIFF
--- a/src/list.rs
+++ b/src/list.rs
@@ -93,11 +93,11 @@ impl List {
     /// Inserts a new entry into the list.
     pub(crate) fn insert(&mut self, entry: Entry) -> NonZeroUsize {
         // Replace the tail with the new entry.
-        let key = NonZeroUsize::new(self.entries.vacant_key());
-        match mem::replace(&mut self.tail, key) {
-            None => self.head = key,
+        let key = NonZeroUsize::new(self.entries.vacant_key()).unwrap();
+        match mem::replace(&mut self.tail, Some(key)) {
+            None => self.head = Some(key),
             Some(t) => {
-                self.entries[t.get()].next.set(key);
+                self.entries[t.get()].next.set(Some(key));
                 entry.prev.set(Some(t));
             }
         }
@@ -111,7 +111,7 @@ impl List {
         self.entries.insert(entry);
 
         // Return the key.
-        key.unwrap()
+        key
     }
 
     /// Removes an entry from the list and returns its state.
--- a/src/list.rs
+++ b/src/list.rs
@@ -1,13 +1,12 @@
 //! The inner list of listeners.
 
-use crate::sync::atomic::{AtomicUsize, Ordering};
 use crate::sync::cell::Cell;
 use crate::Task;
 
-use alloc::boxed::Box;
-
 use core::mem;
-use core::ptr::{self, NonNull};
+use core::num::NonZeroUsize;
+
+use slab::Slab;
 
 /// The state of a listener.
 pub(crate) enum State {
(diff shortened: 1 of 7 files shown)


src/queue.rs
```diff
@@ -37,7 +37,6 @@ impl Queue {
 
     /// Push a node to the tail end of the queue.
     pub(crate) fn push(&self, node: Node) {
-        node.enqueue();
```
zseri, line 40: what did this previously?

src/node.rs
```diff
@@ -109,3 +99,24 @@ impl Node {
...
+    ///
+    /// Returns `Some` with the entry ID if we are no longer queued.
+    pub(crate) fn status(&self) -> Option<NonZeroUsize> {
+        NonZeroUsize::new(self.entry_id.load(Ordering::Acquire))
+    }
+
+    /// Register a listener.
+    pub(crate) fn register(&self, task: Task) {
+        // Set the task.
+        self.task.store(Some(task));
```
zseri, line 114: I guess potentially overwriting the currently stored task is ok... why?

zseri (commented):
looks overall ok

zseri:
I noticed we also still have a `Mutex` impl in the codebase... the reason it is here instead of in `async-lock` or such is because `async-lock` depends on `event-listener`, right?

notgull:
Yes, although the mutex here uses a spinning strategy that we don't expose for that reason. Truth be told, I was thinking of just using the Mutex from the `spin` crate here instead to reduce code deduplication.

src/node.rs
notgull: I don't think it is, actually. I wake it up now.

src/queue.rs
notgull: This toggled a flag in the node that enqueued it, but that flag is no longer used.

src/list.rs
```diff
@@ -36,145 +35,117 @@ impl State {
...
-    pub(crate) fn insert(&mut self, entry: NonNull<Entry>) {
+    pub(crate) fn insert(&mut self, entry: Entry) -> NonZeroUsize {
         // Replace the tail with the new entry.
-        match mem::replace(&mut self.tail, Some(entry)) {
-            None => self.head = Some(entry),
-            Some(t) => unsafe {
-                t.as_ref().next.set(Some(entry));
-                entry.as_ref().prev.set(Some(t));
-            },
+        let key = NonZeroUsize::new(self.entries.vacant_key());
```
zseri, line 96: unwrap the key here, it would make more sense

zseri (approved):

notgull:
With some finagling, I've managed to reduce the scope of the `unsafe` blocks further.

image

PR #40, opened 2022-11-21
Add a "tag" to the event

notgull:
#13 seems like a good idea, so I've recreated it with the newer setup.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -721,7 +721,11 @@ impl<T: Clone> EventListener<T> {
                         // Remove the entry and check if notified.
                         let mut list = lock();
                         let state = list.remove(entry);
-                        return state.notified();
+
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -428,24 +354,146 @@ impl Event {
...
+                        tag: value.clone(),
+                    }));
+                }
+            }
+        }
+    }
+
+    /// Notifies a number of active and still unnotified listeners with a value without emitting a
+    /// `SeqCst` fence.
+    pub fn notify_additional_tagged_relaxed(&self, n: usize, value: &T) {
```
notgull, line 475: Definitely open to bikeshedding the names like this.

notgull:
Needs a significant rework.

PR #41, opened 2022-11-21
Replace custom spinlock with spin

notgull:
This deduplicates the custom spinlock we have with the one that's currently used in the `spin` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ __test = []
 crossbeam-utils = { version = "0.8.12", default-features = false }
 parking = { version = "2.0.0", optional = true }
 slab = { version = "0.4.7", default-features = false }
+spin = { version = "0.9.4", default-features = false, features = ["spin_mutex"] }
 
 [dev-dependencies]
 waker-fn = "1"
--- a/src/inner.rs
+++ b/src/inner.rs
@@ -3,8 +3,7 @@
 use crate::list::List;
 use crate::node::Node;
 use crate::queue::Queue;
-use crate::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use crate::sync::cell::UnsafeCell;
+use crate::sync::atomic::{AtomicUsize, Ordering};
 use crate::Task;
 
 use alloc::vec;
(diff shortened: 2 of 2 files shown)


taiki-e:
Hmm, why do we have spinlock? Sorry for not being able to review your recent PRs, but I remember I disagreed with spinlock in a no-std related issue in this repository -- AFAIK, using spinlock around pure ops like arithmetic operations is fine, but using it around allocations is not always safe (may cause a deadlock). In fact, there is [a report](https://github.com/crossbeam-rs/crossbeam/issues/821#issuecomment-1321261879) that the spinlock around the allocations at crossbeam-channel (recently removed) is believed to have caused a deadlock.

taiki-e:
Ah, it seems the current implementation doesn't have a problem with deadlock as it is try_lock only and the spinning is not unbounded.

notgull:
Yes, the main idea is that it uses a spin lock in the case where it's theoretically ideal and falls back to an atomic queue during periods of high contention. I wrote a more detailed write-up in #34.

(Also, don't apologize for not being able to review my PRs. I know you're busy, and your well-being is more important than my code. 🙂)

notgull:
Note than `spin`'s master branch supports our MSRV as of mvdnes/spin-rs#131.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ __test = []
 crossbeam-utils = { version = "0.8.12", default-features = false }
 parking = { version = "2.0.0", optional = true }
 slab = { version = "0.4.7", default-features = false }
+spin = { version = "0.9.4", default-features = false, features = ["spin_mutex"] }
```
fogti, line 28: imo it would be more readable to replace this with
```toml
[dependencies.spin]
version = "0.9.4"
default-features = false
features = ["spin_mutex"]
```

src/inner.rs
```diff
@@ -186,28 +171,19 @@ impl<T> Mutex<T> {
...
 
 struct MutexGuard<'a, T> {
-    mutex: &'a Mutex<T>,
-}
-
-impl<'a, T> Drop for MutexGuard<'a, T> {
-    fn drop(&mut self) {
-        self.mutex.locked.store(false, Ordering::Release);
-    }
+    guard: spin::MutexGuard<'a, T>,
```
fogti, line 174: wouldn't it make more sense to just do `use spin::MutexGuard;` here`

fogti (changes requested):

notgull:
Taking another look at this, I'm not a fan of incurring the extra dependency.

PR #42, opened 2022-11-22
Experimental no-alloc implementation

notgull:
This PR reimplements `event-listener` using a strategy that does not allocate. This would be suitable for platforms that do not support allocation (e.g. embedded systems).

## Pros

This would allow `event-listener` and its subcrates (e.g. `async-lock`) to be used in nearly every environment that Rust can compile to. With the `portable-atomic` crate, it could even run on single-core systems without atomics.

In addition, by removing allocation, we've more than doubled the speed of this crate in benchmarks.

image

## Cons

This change requires a breaking change in the API. `EventListener` needs to be pinned now, which means that all current users of the `event-listener` crate will have to rewrite their crates to use pinning.

This change also involves using a raw spinlock to protect the operations. While it's only a few arithmetic operations, it may cause deadlocks on embedded systems if, e.g, an interrupt is called during the critical section. I'm not actually sure about the semantics of this. See my "test" below.

A potential solution to this would be to also keep a version of the current 'backup queue' setup for when the mutex fails. This would ensure that spinning is never unbounded in no use case, at the cost of needing the `alloc` crate. However, since the critical section is just a few arithmetic operations, this may not be needed.

This change increases the size of the `Event` and `EventListener` structures, which may cause slowdowns. We could just `Box` it in downstream crates (e.g. `async-channel`, `async-lock`) but that feels counterproductive.

This change introduces a large amount of unsafe code to this crate.

The discussion question is, is this the direction we want to move forwards with this crate?


notgull:
I found an old Arduino kit in my garage, so I decided to give this new code a try on it.

bred board

[Here is the code I used](https://github.com/notgull/evl-arduino-test). Basically, hitting the button causes an interrupt that notifies the main loop blocking on an `Event`. I couldn't get it trapped inside of the spin loop of the `Event` no matter what I tried, so I think it's good to go?

notgull:
@smol-rs/admins I think that we should discuss this, since it concerns the overall future direction of `event-listener`.

zeenix:
@notgull Nice work! The picture of your code in action was a nice :cherries: on top. :smile: 

The only concern would be the API break and it's a pretty big one unfortunately. I for one am fine with that if we bump the semver appropriately. But if we do that, perhaps we should look into the code and get any other breakages we need in the same version bump.

As for the `unsafe` usage, I don't think you added  lot and most of it seems quite safe.

notgull:
I've soured on this PR somewhat. I'm still scared of the spinlock here, even if it only protects arithmetic operations; it's a ticking time bomb waiting to go off. I'll just close this for now.

PR #48, opened 2023-03-14
feat: Add an "is_notified" function

notgull:
This PR adds an `is_notified()` function that tells if any listeners are currently notified.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -205,6 +205,26 @@ impl<T> Event<T> {
         }
     }
 
+    /// Tell whether any listeners are currently notified.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use event_listener::Event;
+    ///
+    /// let event = Event::new();
+    /// let listener = event.listen();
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -191,6 +191,26 @@ impl Event {
...
+    /// let listener = event.listen();
+    /// assert!(!event.is_notified());
+    ///
+    /// event.notify(1);
+    /// assert!(event.is_notified());
+    /// ```
+    #[inline]
+    pub fn is_notified(&self) -> bool {
+        self.try_inner()
+            .map_or(false, |inner| inner.notified.load(Ordering::Relaxed) > 0)
```
fogti, line 211: is `Relaxed` really the correct ordering for this?

notgull: `Acquire` might be better, given that a `Release` ordering is used to store the `notified` atomic.

fogti: yep

fogti (approved):

PR #49, opened 2023-03-14
feat: Add the event-listener-strategy crate

notgull:
See [here](https://github.com/smol-rs/async-lock/pull/39#discussion_r1123845355). This extracts the strategy used for blocking/nonblocking in `async-lock` and `async-channel` into its own crate.

### DIFF
--- a/strategy/Cargo.toml
+++ b/strategy/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "event-listener-strategy"
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+event-listener = { path = "../", default-features = false }
+pin-project-lite = "0.2.9"
+pin-utils = "0.1.0"
+
+[features]
+default = ["std"]
+std = ["event-listener/std"]
+
+[dev-dependencies]
+futures-lite = "1.12.0"
--- a/strategy/src/lib.rs
+++ b/strategy/src/lib.rs
@@ -1,0 +1,391 @@
+// SPDX-Licenser-Identifier: MIT OR Apache-2.0
+
+//! A strategy for using the [`event-listener`] crate in both blocking
+//! and non-blocking contexts.
+//!
+//! One of the stand-out features of the [`event-listener`] crate is
+//! the ability to use it in both asynchronous and synchronous contexts.
+//! However, sometimes using it like this causes a lot of boilerplate
+//! to be duplicated. This crate aims to reduce that boilerplate by
+//! providing an [`EventListenerFuture`] trait that implements
+//! both blocking and non-blocking functionality.
+//!
+//! # Examples
+//!
+//! ```
+//! use event_listener::{Event, EventListener};
+//! use event_listener_strategy::{EventListenerFuture, FutureWrapper, Strategy};
+//!
+//! use std::pin::Pin;
+//! use std::task::Poll;
+//! use std::thread;
+//! use std::sync::Arc;
+//!
+//! // A future that waits three seconds for an event to be fired.
+//! fn wait_three_seconds() -> WaitThreeSeconds {
+//!     let event = Event::new();
+//!     let listener = event.listen();
+//!
+//!     thread::spawn(move || {
+//!         thread::sleep(std::time::Duration::from_secs(3));
+//!         event.notify(1);
+//!     });
+//!
+//!     WaitThreeSeconds { listener: Some(listener) }
+//! }
+//!
+//! struct WaitThreeSeconds {
+//!     listener: Option<EventListener>,
+//! }
+//!
+//! impl EventListenerFuture for WaitThreeSeconds {
+//!     type Output = ();
+//!
+//!     fn poll_with_strategy<S: Strategy>(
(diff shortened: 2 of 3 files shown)


strategy/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "event-listener-strategy"
+version = "0.1.0"
+edition = "2018"
```
zeenix, line 4: Can we just go with 2021 maybe? or do we want it consistent across all smol-rs crates?

notgull: I want to use this in `async-lock` and `async-channel`, so it needs to be 2018 edition.

strategy/src/lib.rs
```diff
@@ -0,0 +1,391 @@
+// SPDX-Licenser-Identifier: MIT OR Apache-2.0
+
+//! A strategy for using the [`event-listener`] crate in both blocking
```
zeenix, line 3: nit: the limit is 100 chars (74 is for git commits :)) and best to make use of it to reduce LOCs.

strategy/Cargo.toml
zeenix: Edition is strictly specific to the crate itself, doesn't affect the dependent crates.

fogti: what's the implied MSRV of edition 2021?

strategy/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[dependencies]
+event-listener = { path = "../", default-features = false }
```
fogti, line 7: I think a version should be specified for `event-listener`.

strategy/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+/// ```
+#[macro_export]
+macro_rules! easy_wrapper {
+    (
+        $(#[$meta:meta])*
+        $vis:vis struct $name:ident ($inner:ty => $output:ty);
+        $(#[$wait_meta:meta])*
+        $wait_vis: vis wait();
+    ) => {
+        ::pin_project_lite::pin_project! {
```
fogti, line 145: this requires a dependency on `pin-project-lite` at the usage site, is that intended? (alternative would be re-exporting `pin_project_lite` in this crate (perhaps with `#[doc(hidden)]`), and prefixing this with `$crate`

strategy/src/lib.rs
```diff
@@ -0,0 +1,407 @@
...
+            $wait_vis fn wait(self) -> $output {
+                use $crate::EventListenerFuture;
+                self._inner.into_inner().wait()
+            }
+        }
+
+        impl ::core::future::Future for $name {
+            type Output = $output;
+
+            fn poll(
```
fogti, line 170: it might be a good idea to mark this with `#[inline]`

fogti:
I'd sprinkle `#[inline]` on basically all `pub` one-line-functions.

strategy/src/lib.rs
zeenix: I've always wondered if there is any emperical evidence of that helping anything. :thinking: Aren't the compilers these days smart enough to handle that, but with the added advantage of keeping the user-selected optimization level in mind?

strategy/Cargo.toml
zeenix: [1.56](https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html) apparently.

strategy/src/lib.rs
fogti: using `#[inline]` should just act as a hint, so it shouldn't massively hurt... I'm not how big the effect would be, benchmarking could sort-of prove it's effect...

strategy/Cargo.toml
zeenix: hmm.. that would be problematic cause MSRV of event-listener itself is 1.39. So nm my original suggestion here but we should declare our MSRV in this new crate to ensure it's not implicitly incompatible (or accidently becomes so in the future).

strategy/src/lib.rs
zeenix: I know what they are intended for. They **do** hurt readability a bit so if there is no empirical evidence of them actually being useful, I would rather not have them.

zeenix: I asked the internets and the call was heeded: https://toot.cat/@jamey/110023643339747938 Reading the relevant part of that Rust book pointed there, and seeing it's a oneliner, I think you're right. This should be inlined across crates.

Sorry for the noise. I was mainly just wanting to make sure that extra line is worth it.

notgull: Okay, I've inlined most of the small functions in this crate.

fogti (approved):

zeenix (approved):

fogti:
oh, I also have some questions about the usage of `feature = "std"`.
it seems to be missing here:
https://github.com/smol-rs/event-listener/blob/52cb4bfd1afcae1867a8fe8ba215394958e8b86f/strategy/src/lib.rs#L420

PR #51, opened 2023-03-16
feat: Rework the internal structure of the event-listener crate

notgull:
Alright, here we go again.

This PR once again reworks the internal structure of `event-listener`. An intrusive linked list is the fastest way to handle an event queue, but isn't available in libstd. Therefore, this PR reworks the internals such that the implementation of event handling used depends on whether or not the libstd feature is enabled:

- If `std` is enabled, then it uses the intrusive linked list.
- If `std` isn't enabled, then it falls back to the current `master` strategy of using a linked list arena falling back to an atomic queue.

I'm pretty happy with this iteration at this stage. Draft because I still need to refine the tests.


benches/bench.rs
```diff
@@ -1,22 +1,31 @@
...
 
 fn bench_events(c: &mut Criterion) {
     c.bench_function("notify_and_wait", |b| {
         let ev = Event::new();
-        b.iter(|| {
-            let mut handles = Vec::with_capacity(COUNT);
+        let mut handles = Vec::with_capacity(COUNT);
+        for _ in 0..COUNT {
+            handles.push(EventListener::new(&ev));
+        }
```
fogti, line 14:
```suggestion
        let mut handles = core::iter::repeat_with(|| EventListener::new(&ev))
            .take(COUNT)
            .collect::<Vec<_>>();
```

fogti:
ok, so some changes do affect lifetimes and sprinkle `as_mut()` all over the place, can we split them out as much as possible (perhaps even do them without changing any functionaliy, e.g. via `PhantomData` for lifetimes, so that the noise is reduced)? Reviewing this in the current state is really hard, as it is not clear what is moved and how.

src/std.rs
```diff
@@ -188,7 +188,7 @@ impl Inner {
         match next {
             None => self.tail = prev,
             Some(n) => unsafe {
-                n.as_ref().next.set(prev);
+                n.as_ref().prev.set(prev);
```
fogti, line 191: what happens here?

fogti:
* The commit `fmt` should be squashed with the commit which introduced the formatting divergence (that is, modified that code segment previously).
* The commit `Wait tests are hanging on MIRI, not sure why` which seems to affect the first commit should also be squashed into it.
* The second commit `Build ndf into C/I` appears ok, but the commit message should be refined to explain it better.
* The first commit `Incomplete` should then be split into several, optimally independent commits which should compile on their own and demonstrate what code is moved. This branch should not be squashed upon merging, as the structure and hopefully step-wise natrue of the resulting commits is by then probably necessary to understand what is going on.
* The last commit might be further refined, but also appears currently ok

src/std.rs
notgull: There was a bug in there that was causing the MIRI issues. It should be fixed now.

src/lib.rs
```diff
@@ -120,6 +126,42 @@ impl Task {
...
+    /// Tells if this task will wake up the other task.
+    fn will_wake(self, other: Self) -> bool {
+        match (self, other) {
+            (Self::Waker(a), Self::Waker(b)) => a.will_wake(b),
+            #[cfg(feature = "std")]
+            (Self::Unparker(_), Self::Unparker(_)) => {
+                // TODO: Use unreleased will_unpark API.
+                false
+            }
+            #[cfg(feature = "std")]
```
fogti: tying this to `"std"` is probably not a good idea. I'd prefer using `#[allow(unused)]` instead or making all cases explicit.

src/list.rs
```diff
@@ -339,64 +364,36 @@ impl ListenerSlab {
...
+                return;
             }
+            n -= self.notified;
         }
-    }
 
-    /// Notifies a number of additional entries.
-    #[cold]
-    pub(crate) fn notify_additional(&mut self, mut n: usize) {
         while n > 0 {
```
fogti: (maybe later) a `for` loop would be nicer here (but not necessary, so only if it wouldn't occur a performance penality (and that it hasn't been done yet probably indicates that it does incur a penality idk; if that gets tested, it warrants a comment))

src/list.rs
```diff
@@ -14,6 +14,14 @@ use core::ops;
 use slab::Slab;
 
 impl crate::Inner {
+    /// Locks the list.
+    fn lock(&self) -> Option<ListGuard<'_>> {
```
fogti: this method ought to be named `try_lock`, right?

src/list.rs
```diff
@@ -1,7 +1,14 @@
 //! The inner list of listeners.
 
-use crate::node::{Node, TaskWaiting};
-use crate::queue::Queue;
+#[path = "list/node.rs"]
```
fogti: why are here paths specified explicitly?

src/list.rs
```diff
@@ -248,21 +249,140 @@ impl Drop for ListGuard<'_> {
...
+impl fmt::Debug for Entry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        struct TakenState<'a> {
+            state: Option<State>,
+            cell: &'a Cell<State>,
+        }
+
+        impl Drop for TakenState<'_> {
+            fn drop(&mut self) {
+                self.cell.set(self.state.take().unwrap());
```
fogti: hmm I think it would be nicer to reuse a kind of `State` here as a sentinel/placeholder, e.g. `core::mem::replace(&mut self.state, State::NotifiedTaken)` or such.b

src/list.rs
```diff
@@ -248,21 +249,140 @@ impl Drop for ListGuard<'_> {
...
+            }
+            Entry::Empty(next) => f.debug_tuple("Empty").field(next).finish(),
+            Entry::Sentinel => f.debug_tuple("Sentinel").finish(),
+        }
+    }
+}
+
+impl PartialEq for Entry {
+    fn eq(&self, other: &Entry) -> bool {
+        struct RestoreState<'a> {
```
fogti: this code duplication feels unnecessary. (compare to the above `TakenState`)

src/std.rs
```diff
@@ -0,0 +1,344 @@
...
+                Some(true)
+            }
+
+            State::Task(other_task) => {
+                // Only replace the task if it's different.
+                if !task.will_wake(other_task.as_task_ref()) {
+                    entry.state.set(State::Task(task.into_task()));
+                } else {
+                    entry.state.set(State::Task(other_task));
+                }
```
fogti, line 148: these branches should be condensed (`entry.state.set(State::Task(if !... ))`)

fogti:
thanks, looks overall LGTM.

src/list.rs
notgull: I renamed it to `try_lock` in a later commit

fogti (approved):

PR #52, opened 2023-04-03
feat: Add tagged notifications

notgull:
Supersedes #40 by adding tags to the event. Similarly to #40, this PR:

- Adds a `T` parameter to `Event` and `EventListener` that is `()` by default.
- **Breaking:** Makes it so `wait()` and friends return `T` or `Option<T>` instead of `()` and `bool`.
- Makes it so polling `EventListener` as a future returns `T` instead of `()`.

Unlike #40, I wanted to use a more sustainable method for notifying the event. Introducing the **Notification** trait. Its API looks like this:

```rust
trait Notification {
    type Tag;
    fn is_additional(&self) -> bool;
    fn count(&self) -> usize;
    fn fence(&self);
    fn next_tag(&mut self) -> T;
}
```

This roughly corresponds to the implicit parameters to the current `event.notify_<is_additional>_<relaxed>(count)` function. `is_additional` sets the `additional` property to `true`, and `relaxed` makes the fence do nothing instead of emitting a `SeqCst` fence. The `next_tag` function emits the tag that gets attached to the notification.

I wanted to give this some power, so I created the **IntoNotification** trait as well.

```rust
trait IntoNotification {
    type Tag;
    type Notify: Notification<Tag = Self::Tag>;
    fn into_notification(self) -> Self::Notify;
    
    fn additional(self) -> Additional<Self::Notify> { .. }
    fn relaxed(self) -> Relaxed<Self::Notify> { .. }
    fn tag<T: Clone>(self, tag: T) -> Tag<Self::Notify, T> { .. }
    fn tag_with<T, F: FnMut() -> T>(self, f: F) -> TagWith<Self::Notify, F> { .. }
}

impl<N: Notification> IntoNotification for N { .. }
impl IntoNotification for usize { .. }
```

This way, if we make `notify()` take an `impl IntoNotification` instead of `usize`, it still works as intended, as the implementation for `usize::into_notification` is non-additional and non-relaxed:

```rust
event.notify(3);
```

You can also use this system to add combinators (`.additional()`, `.relaxed()`), such that:

```rust
// Old code:
event.notify_additional(3);

// New code:
event.notify(3.additional());
```

Then, to add tags to a notification, you would do:

```rust
// Assume tag is a Vec<usize>
event.notify(3.tag(vec![1, 2, 3]));

// Alternate method
event.notify(3.tag_with(|| vec![1, 2, 3]));
```

Discussion questions:

- Should `Notification` and `IntoNotification` be sealed? If it is, it means we can add new methods to it without worrying. If it isn't, it means that users can implement their own notification types. However, that may cause behavior we don't expect.

### DIFF
--- a/src/no_std/node.rs
+++ b/src/no_std/node.rs
@@ -52,7 +52,7 @@ pub(crate) struct TaskWaiting {
     entry_id: AtomicUsize,
 }
 
-impl<T: Unpin> Node<T> {
+impl<T> Node<T> {
     pub(crate) fn listener() -> (Self, Arc<TaskWaiting>) {
         // Create a new `TaskWaiting` structure.
         let task_waiting = Arc::new(TaskWaiting {
--- a/src/no_std/node.rs
+++ b/src/no_std/node.rs
@@ -2,6 +2,7 @@
 
 //! The node that makes up queues.
 
+use crate::notify::GenericNotify;
 use crate::sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
 use crate::sync::Arc;
 use crate::sys::ListenerSlab;
@@ -12,8 +13,10 @@ use alloc::boxed::Box;
 use core::num::NonZeroUsize;
 use core::ptr;
 
+pub(super) type GenericTags<T> = Box<dyn FnMut() -> T + Send + Sync + 'static>;
+
 /// A node in the backup queue.
-pub(crate) enum Node {
+pub(crate) enum Node<T> {
     /// This node is requesting to add a listener.
     // For some reason, the MSRV build says this variant is never constructed.
     #[allow(dead_code)]
@@ -23,13 +26,7 @@ pub(crate) enum Node {
     },
 
     /// This node is notifying a listener.
-    Notify {
-        /// The number of listeners to notify.
-        count: usize,
-
-        /// Whether to wake up notified listeners.
-        additional: bool,
-    },
+    Notify(GenericNotify<GenericTags<T>>),
 
     /// This node is removing a listener.
     RemoveListener {
@@ -55,7 +52,7 @@ pub(crate) struct TaskWaiting {
     entry_id: AtomicUsize,
 }
 
-impl Node {
+impl<T: Unpin> Node<T> {
     pub(crate) fn listener() -> (Self, Arc<TaskWaiting>) {
         // Create a new `TaskWaiting` structure.
         let task_waiting = Arc::new(TaskWaiting {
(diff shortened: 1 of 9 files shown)


notgull:
Blocked on taiki-e/portable-atomic#93, since `portable_atomic_util::Arc` doesn't implement `Unpin` and thus can't be projected until it does.

fogti:
> Should Notification and `IntoNotification` be sealed? If it's not, it means we can add new methods to it without worrying. If it is, it means that users can implement their own notification types. However, that may cause behavior we don't expect.

I think you've messed up the negation of "sealed" here.

Also, if we don't seal it, it might be a good idea to split up the `IntoNotification` trait up, so that the functions which users should be able to customize are separated from the rest (where customization wouldn't make sense).

notgull:
> I think you've messed up the negation of "sealed" here.
> 
> Also, if we don't seal it, it might be a good idea to split up the `IntoNotification` trait up, so that the functions which users should be able to customize are separated from the rest (where customization wouldn't make sense).

Thanks for the catch!

I've decided to seal it anyways, since you can do whatever you want using the combinators.

src/no_std/node.rs
```diff
@@ -12,8 +13,10 @@ use alloc::boxed::Box;
 use core::num::NonZeroUsize;
 use core::ptr;
 
+pub(super) type GenericTags<T> = Box<dyn FnMut() -> T + Send + Sync + 'static>;
```
fogti, line 16: is it really a good idea to require an allocation for usage? (why not make it generic in such a way that T is a function itself, which would be more general and probably also faster)

notgull: Unfortunately it's necessary. This structure is stored directly inside of `Queue<T>`, so we can't use the function as a parameter without it propagating all the way to `Event` and `EventListener`.

In addition, this is on a cold path, so it's unlikely to happen in the first place.

fogti: hm so why can't we parameterize `Event` and `EventListener` with that anyways and create type aliases for this case? It would blow up the types a bit, but I don't think it would be bad in practical application

notgull: The closure stored in these boxes isn't a user provided closure; it's [this closure](https://github.com/smol-rs/event-listener/blob/a8c3aa76c2d4aee7f8dd0fa9c2098e753113a917/src/no_std.rs#L140). Until we have TAIT, we can't actually get a type for this closure, therefore we have to box it for now.

fogti: Ohhh... but why do we have to use such a closure and not package it into a custom struct and idk... have it implement `Iterator` or such (or why not just a counter and a generating function?)

notgull: You're right. I've replaced the boxed closure with a `Vec` of tags. It should have the same effect.

notgull:
The MSRV failure looks spurious to me; running `cargo +1.39 build` and `cargo +1.39 build --no-default-features` both work on my machine.

fogti:
I'm not like 100% sure about the design, but it looks ok to me now.

fogti (approved):

PR #57, opened 2023-05-14
Make the notify() function return the number of notified listeners

### LINKED ISSUE #6: [Feature] Getting the number of notified listeners back
Hello!

I've just built an RWLock based around a single atomic and now I'm trying to make an async version with eventlistener. For my purposes, simply knowing if any listeners were notified would suffice, but knowing the count seems a logical extension of that.

What do you think?

notgull:
Closes #6

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -256,6 +256,20 @@ impl<T> Event<T> {
     ///
     /// In certain cases, this function emits a `SeqCst` fence before notifying listeners.
     ///
+    /// This function returns the number of [`EventListener`]s that were notified by this call.
+    ///
+    /// # Caveats
+    ///
+    /// If the `std` feature is disabled, the notification will be delayed under high contention,
+    /// such as when another thread is taking a while to `notify` the event. In this circumstance,
+    /// this function will return `0` instead of the number of listeners actually notified. Therefore
+    /// if the `std` feature is disabled the return value of this function should not be relied upon
+    /// for soundness and should be used only as a hint.
+    ///
+    /// If the `std` feature is enabled, no spurious returns are possible, since the `std`
+    /// implementation uses system locking primitives to ensure there is no unavoidable
+    /// contention.
+    ///
     /// # Examples
     ///
     /// Use the default notification strategy:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -360,7 +360,7 @@ impl<T> Event<T> {
     /// event.notify(1.additional().relaxed());
     /// ```
     #[inline]
-    pub fn notify(&self, notify: impl IntoNotification<Tag = T>) {
+    pub fn notify(&self, notify: impl IntoNotification<Tag = T>) -> usize {
         let notify = notify.into_notification();
 
         // Make sure the notification comes after whatever triggered it.
@@ -376,9 +376,11 @@ impl<T> Event<T> {
             // Notify if there is at least one unnotified listener and the number of notified
             // listeners is less than `limit`.
             if inner.notified.load(Ordering::Acquire) < limit {
-                inner.notify(notify);
+                return inner.notify(notify);
             }
         }
+
+        0
     }
 
     /// Return a reference to the inner state if it has been initialized.
@@ -489,7 +491,7 @@ impl Event<()> {
     /// event.notify_relaxed(2);
     /// ```
     #[inline]
-    pub fn notify_relaxed(&self, n: usize) {
+    pub fn notify_relaxed(&self, n: usize) -> usize {
         self.notify(n.relaxed())
     }
 
@@ -538,7 +540,7 @@ impl Event<()> {
     /// event.notify_additional(1);
     /// ```
     #[inline]
-    pub fn notify_additional(&self, n: usize) {
+    pub fn notify_additional(&self, n: usize) -> usize {
         self.notify(n.additional())
     }
 
@@ -592,7 +594,7 @@ impl Event<()> {
     /// event.notify_additional_relaxed(1);
     /// ```
     #[inline]
-    pub fn notify_additional_relaxed(&self, n: usize) {
+    pub fn notify_additional_relaxed(&self, n: usize) -> usize {
         self.notify(n.additional().relaxed())
     }
 }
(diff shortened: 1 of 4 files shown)


src/no_std.rs
```diff
@@ -576,7 +576,7 @@ impl<T> ListenerSlab<T> {
 
             // Notify the next entry.
             match self.start {
-                None => break,
+                None => return original_count - n - 1,
```
zeenix, line 579: The "Fix off-by-one error in implementation" commit seems to be a fixup commit for the first one. If that's the case, please squash it into that and force push. I am a champion of atomic commits but that's more about each commit keeping only one logical commit rather than immutability of the commits.

src/lib.rs
```diff
@@ -234,6 +234,12 @@ impl<T> Event<T> {
     /// The default implementation (implemented on `usize`) is a notification that only notifies
     /// *at least* the specified number of listeners.
     ///
+    /// This function returns the number of [`EventListener`]s that were notified by this
+    /// call. However, in some cases, the notification will be delayed and this function will
+    /// return `0` instead of the number of listeners actually notified. Therefore, the return
+    /// value of this function should not be relied upon for soundness and should be used only
+    /// as a hint.
```
zeenix, line 241: This sounds very fishy to me and will basically just mean "I best not use this" if they can't rely on it. At the very least, we should specify the conditions under which the value is unreliable so users have some clue of when they can and can not rely on it.

zeenix (changes requested):

src/lib.rs
notgull: I've specified how it works under the hood now

src/lib.rs
```diff
@@ -256,6 +256,21 @@ impl<T> Event<T> {
...
+    ///
+    /// This function returns the number of [`EventListener`]s that were notified by this
+    /// call. However, in some cases, the notification will be delayed and this function will
+    /// return `0` instead of the number of listeners actually notified. Therefore, the return
+    /// value of this function should not be relied upon for soundness and should be used only
+    /// as a hint.
+    ///
+    /// The cases where `0` will be spuriously returned include:
+    ///
+    /// - If the `std` feature is enabled, no spurious returns are possible, since the `std`
```
zeenix, line 269: sorry to nitpick but that's a strange way of putting it. You say you're going to list cases where `0` will be returned and then the first item is a case where `0` will not be returned.

zeenix: LGTM otherwise. :+1:

src/lib.rs
```diff
@@ -256,6 +256,22 @@ impl<T> Event<T> {
...
+    /// value of this function should not be relied upon for soundness and should be used only
+    /// as a hint.
+    ///
+    /// The cases where `0` will be spuriously returned include:
+    ///
+    /// - If the `std` feature is disabled, spurious returns will happen under high contention,
+    ///   such as when another thread is taking a while to `notify` the event.
+    /// - If the `std` feature is enabled, no spurious returns are possible, since the `std`
+    ///   implementation uses system locking primitives to ensure there is no unavoidable
+    ///   contention.
```
zeenix, line 273: Maybe I wasn't very clear, sorry. This list item is about when spurious return is impossible when this is supposed to be list of cases of spurious returns. Looks to me that we don't need a list here at all. Maybe just convert the whole section into paragraphs?

notgull: Okay, I converted the list into a paragraph

src/lib.rs
```diff
@@ -256,6 +256,19 @@ impl<T> Event<T> {
...
+    /// This function returns the number of [`EventListener`]s that were notified by this
+    /// call. However, in some cases, the notification will be delayed and this function will
+    /// return `0` instead of the number of listeners actually notified. Therefore, the return
+    /// value of this function should not be relied upon for soundness and should be used only
+    /// as a hint.
+    ///
+    /// If the `std` feature is disabled, spurious returns will happen under high contention,
+    /// such as when another thread is taking a while to `notify` the event. If the `std` 
+    /// feature is enabled, no spurious returns are possible, since the `std` implementation 
+    /// uses system locking primitives to ensure there is no unavoidable contention.
```
zeenix, line 270: Thanks. I still feel this can be clearer. People don't always read carefully. They'll only see the first paragraph and assume that this value is just in general unreliable. So how about something like this:

```suggestion
    /// This function returns the number of [`EventListener`]s that were notified by this call.
    ///
    /// #Caveats
    ///
    /// If the `std` feature is disabled, the notification will be delayed under high contention,
    /// such as when another thread is taking a while to `notify` the event,  and this function
    /// will return `0` instead of the number of listeners actually notified. Therefore, if the `std`
    /// feature is disabled the return value of this function should not be relied upon for
    /// soundness and should be used only as a hint.
    ///
    /// If the `std`  feature is enabled, no spurious returns are possible, since the `std`
    /// implementation uses system locking primitives to ensure there is no unavoidable
    /// contention.
```

English isn't my first language though so feel free to modify/improve the wording.

zeenix (changes requested):

src/lib.rs
notgull: Sounds good to me! I edited it a little and pushed it.

zeenix (approved):

zeenix:
Sorry to bring this back on the table but I keep wondering if there is a use case for returning a number that can't be relied on? I can imagine this difference ending up with a disaster scenario: I start using this API and relying on the return value (let's say a lot) because I'm using `std` today. Down the line, I start supporting nostd and disable `std` for that. I seriously doubt I'll remember this caveat and continue depending on the return value.

Hence, unless there is a compelling use-case for it, I think it'd be better if the method just did not return the number of listeners if `std` is disabled (and we document this).

notgull:
That's a fair point. This is a massive footgun waiting to happen. Maybe we set up this method to return a `NotifiedListeners` type, that has a method returning a `usize` only on `std`?

zeenix:
> That's a fair point. This is a massive footgun waiting to happen.

Glad you agree.



> Maybe we set up this method to return a `NotifiedListeners` type, that has a method returning a `usize` only on `std`?

Not sure what would be the advantage of that over providing different signatures for different features. It will only be an indirection.

PR #67, opened 2023-08-09
Remove most of Pin API related unsafe code

taiki-e:


Cargo.toml
```diff
@@ -21,6 +21,7 @@ portable-atomic = ["portable-atomic-util", "portable_atomic_crate"]
 
 [dependencies]
 parking = { version = "2.0.0", optional = true }
+pin-project-lite = "0.2.12"
```
taiki-e, line 24: Given that all smol-rs crates that depend on event-listener will depend on pin-project-lite, at least after the update to event-listener v3, I see no reasonable reason to avoid a dependency on pin-project-lite here.

notgull: I see; that makes sense to me, and would be my only real quibble with this PR.

notgull (approved):
Thanks!

PR #68, opened 2023-08-09
Fix typos

taiki-e:


notgull (approved):
Thanks!

It might be worth it to have a spell checker in CI for the `smol-rs` crates.

PR #69, opened 2023-08-12
Notify user if their tag may be dropped

notgull:
Due to the implementation of the no_std backend, it is possible for a tag created by the user to be dropped before it is notified. This makes it  poor idea to use the tag for, e.g. sending data over a channel. Therefore this PR adds a parameter for the with_tag closure that gives the user information on whether or not the tag may be dropped.


notgull:
Superseded by #75

PR #71, opened 2023-08-13
Use inline assembly in full_fence

### LINKED ISSUE #70: UB due to lack of a full fence in full_fence (on x86)
I was recently made aware of this code:

https://github.com/smol-rs/event-listener/blob/0ea464102e74219aab2932f9eff14418a13268d4/src/notify.rs#L562-L577

As the comment says, this is UB -- and the myth that no sane compiler is going to optimize atomics [is a myth](https://github.com/jfbastien/papers/blob/master/source/N4455.rst). However, I admit I don't know enough about this specific case to say whether this is a risk for this crate.

What I am wondering, why doesn't this use inline assembly? That is the intended mechanism to force the compiler to generate code in a particular way, and it is clearly sound here. The alternative is even mentioned in the comment, but unfortunately without an explanation of why it was discarded.

taiki-e:
Fixes #70 

cc @RalfJung


src/notify.rs
```diff
@@ -555,27 +555,31 @@ impl_for_numeric_types! { usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 }
...
+        // It is common to use `lock or` here, but when using a local variable, `lock not`, which
+        // does not change the flag, should be slightly more efficient.
+        // Refs: https://www.felixcloutier.com/x86/not
+        unsafe {
+            #[cfg(target_pointer_width = "64")]
+            asm!("lock not qword ptr [{0}]", in(reg) a.get(), options(nostack, preserves_flags));
+            #[cfg(target_pointer_width = "32")]
+            asm!("lock not dword ptr [{0:e}]", in(reg) a.get(), options(nostack, preserves_flags));
+        }
+        return;
```
notgull, line 579: Do we need `compiler_fence`s here as well?

notgull (commented):
Overall LGTM, but I'll let @RalfJung, the human UB detector, make the final call

src/notify.rs
taiki-e: No. [asm without nomem/readonly implies compiler fence](https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm/topic/asm.20and.20compiler.20fences/near/274118488).

src/notify.rs
```diff
@@ -555,27 +555,31 @@ impl_for_numeric_types! { usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 }
...
-        // temporary atomic variable and compare-and-exchanging its value. No sane compiler to
-        // x86 platforms is going to optimize this away.
-        atomic::compiler_fence(Ordering::SeqCst);
-        let a = AtomicUsize::new(0);
-        let _ = a.compare_exchange(0, 1, Ordering::SeqCst, Ordering::SeqCst);
-        atomic::compiler_fence(Ordering::SeqCst);
-    } else {
+        let a = UnsafeCell::new(0_usize);
+        // It is common to use `lock or` here, but when using a local variable, `lock not`, which
+        // does not change the flag, should be slightly more efficient.
```
taiki-e, line 571: (Note on potential future optimization: I have not benchmarked, but using `xchg` (`lock` prefix is implied) here may potentially be faster than `lock not`. The number of registers used will increase, but the value at `a` and the value to be written can be uninitialized. Making the value at `a` uninitialized with `lock not` is fine from a hardware standpoint, but Valgrind will probably be angry about it.)

RalfJung:
> Overall LGTM, but I'll let @RalfJung, the human UB detector, make the final call

Sorry, when it comes to inline assembly I'm out.^^ I can't even read assembly. ;)

However the compiler certainly has to assume that an inline assembly block (without `nomem` or `readonly`) performs synchronization actions such as an SC fence, so in broad strokes this should be correct, and I trust @taiki-e  to get the details right. :)

notgull (approved):
Thanks!

PR #76, opened 2023-08-17
Tracking down a deadlock in async-channel

notgull:
There's a deadlock that happens infrequently in the `async-channel` tests when `no_std` mode is enabled. This PR enumerates my efforts to track it down.

So far I believe that the issue is that either the listener is not being successfully inserted into the list, or that the notification is somehow not being delivered. Seems to involve inconsistencies brought on by the fallback queue.


notgull:
I've tracked it down now. To quote my commit message:

> There was a deadlock where notifications would not get delivered for
listeners that were stuck in the fallback queue. There is an
optimization where the notification count is checked to see if the
notifications delivered by the user would make any impact on the
listener chain; if it wouldn't, then it isn't done at all. For no_std,
I've removed this optimization, as I doubt that it can be done
consistently due to the nature of the fallback queue.
> 
> During testing I also uncovered problems with the current implementation
of the fallback queue itself. I've replaced it with concurrent-queue for
now, as I doubt that a fully functioning queue wouldn't be trivial
enough to justify inlining it into event-listener itself. Unfortunately
it adds three additional dependencies to the tree of event-listener.
> 
> With this change, the async-channel tests pass consistently.

src/no_std.rs
```diff
@@ -72,40 +83,52 @@ impl<T> crate::Inner<T> {
...
-                        list.remove(key, propagate)
-                    }
+        loop {
+            let state = match listener.as_mut().take() {
+                Some(Listener::HasNode(key)) => {
+                    match self.try_lock() {
+                        Some(mut list) => {
+                            // Fast path removal.
+                            let state = list.remove(key, propagate);
+                            state
```
fogti, line 93: just use
```rust
  list.remove(key, propagate)
```

src/no_std.rs
```diff
@@ -149,19 +175,27 @@ impl<T> crate::Inner<T> {
                     match self.try_lock() {
                         Some(mut guard) => {
                             // Fast path registration.
-                            return guard.register(listener, task);
+                            let result = guard.register(listener, task);
+                            return result;
```
fogti, line 179: this change should be undone (it's a no-op, but bloats the code)

src/no_std.rs
```diff
@@ -210,22 +249,29 @@ pub(crate) struct ListGuard<'a, T> {
...
-        &mut self,
-        start_node: Node<T>,
-        tasks: &mut Vec<Task>,
-        guard: &mut MutexGuard<'_, ListenerSlab<T>>,
-    ) {
+    fn process_nodes_slow(&mut self, start_node: Node<T>) {
         // Process the start node.
-        tasks.extend(start_node.apply(guard));
+        self.tasks
+            .extend(start_node.apply(self.guard.as_mut().unwrap()));
```
fogti, line 262: why can't we hold onto `self.guard.as_mut().unwrap()`? (theoretically, the unnecessary branching should get optimized away, but still, it feels unclean.)

src/no_std.rs
```diff
@@ -270,9 +309,17 @@ impl<T> Drop for ListGuard<'_, T> {
         drop(list);
 
         // Wakeup all tasks.
-        for task in tasks {
+        for task in self.tasks.drain(..) {
             task.wake();
         }
+
+        // There is a deadlock where a node is pushed the end of the queue after we've finished
```
fogti, line 316:
```rust
       // There is a deadlock where a node is pushed to the end of the queue after we've finished
```

src/no_std.rs
```diff
@@ -270,9 +309,17 @@ impl<T> Drop for ListGuard<'_, T> {
...
+        // process_nodes() but before we've finished dropping the lock. This can lead to systems
```
fogti, line 317: systems?

src/no_std.rs
```diff
@@ -270,9 +309,17 @@ impl<T> Drop for ListGuard<'_, T> {
...
+        // halting. Therefore check before we finish dropping if there is anything left in the
+        // queue, and if so, lock it again and force a queue update.
+        if !self.inner.list.queue.is_empty() {
+            self.inner.queue_update();
```
fogti, line 321: imo this should be implemented in a way that it won't recursively call the destructor; instead, it should do so iteratively (e.g. by replacing itself and looping until the queue is empty)

src/no_std.rs
```diff
@@ -632,18 +692,27 @@ impl<T> ListenerSlab<T> {
...
 pub(crate) enum Listener<T> {
     /// The listener has a node inside of the linked list.
     HasNode(NonZeroUsize),
 
     /// The listener has an entry in the queue that may or may not have a task waiting.
     Queued(Arc<TaskWaiting>),
 
-    /// Eat the lifetime for consistency.
+    /// Eat the generic type for consistency.
     _EatLifetime(PhantomData<T>),
```
fogti, line 703: maybe also rename the enum kind accordingly.

src/no_std/node.rs
```diff
@@ -62,14 +69,39 @@ pub(crate) enum Node<T> {
     Waiting(Task),
 }
 
+pub(crate) struct WaitingListener(Arc<TaskWaiting>);
+
+impl<T> fmt::Debug for Node<T> {
```
fogti, line 74: why can't we just implement this on `TaskWaiting` instead?

notgull: This struct was an artifact from when I was trying to track down where the `TaskWaiting` was going. I've now removed it.

src/no_std.rs
notgull: Clarified this

src/no_std.rs
```diff
@@ -292,33 +293,41 @@ impl<T> ops::DerefMut for ListGuard<'_, T> {
 
 impl<T> Drop for ListGuard<'_, T> {
     fn drop(&mut self) {
-        // Process every node left in the queue.
-        self.process_nodes();
+        loop {
```
fogti, line 296:
```suggestion
        while let Some(list) = self.guard.take() {
```

src/no_std.rs
```diff
@@ -292,33 +293,41 @@ impl<T> ops::DerefMut for ListGuard<'_, T> {
...
+            // process_nodes() but before we've finished dropping the lock. This can lead to some
+            // notifications not being properly delivered, or listeners not being added to the list.
+            // Therefore check before we finish dropping if there is anything left in the queue, and
+            // if so, lock it again and force a queue update.
+            if !self.inner.list.queue.is_empty() {
+                self.guard = self.inner.list.inner.try_lock();
+                if self.guard.is_some() {
+                    continue;
+                }
+            }
```
fogti, line 328:
```suggestion
            } else {
                break;
            }
```

fogti:
Looks mostly ok by now (although I'd replace the `.take().unwrap()` and `.is_some()` in the loop with a `while let Some(...)` loop)

notgull:
> Looks mostly ok by now (although I'd replace the `.take().unwrap()` and `.is_some()` in the loop with a `while let Some(...)` loop)

Unfortunately I can't do that exactly, as `self.process_nodes()` involves the guard and the borrow checker doesn't like that. However I did something similar to it.

fogti (approved):

PR #77, opened 2023-08-24
Add explicit web support

notgull:
This commit adds WASM compilation support to this crate. The main thing is that the wait() family of APIs are removed in WASM mode, as blocking is not allowed in WASM.

In addition, tests are added to CI to support WASM.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,11 @@ jobs:
         run: cargo hack build --all --no-dev-deps
       - run: cargo hack build --all --target thumbv7m-none-eabi --no-default-features --no-dev-deps
       - run: cargo hack build --target thumbv7m-none-eabi --no-default-features --no-dev-deps --features portable-atomic
+      - name: Install wasm-pack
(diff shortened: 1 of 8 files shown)


taiki-e:
Thanks!

- Is this also needed for wasi-threads?
- Because of MSRV, we can use `target_family = "wasm"` instead of `any(target_arch = "wasm32", target_arch = "wasm64")`. (`target_family = "wasm"` requires Rust 1.54)
- The changes on tests can be simplified in the same way as https://github.com/smol-rs/fastrand/commit/83f08b49f50db1e434961862cbcb39f9c4c6580a.

notgull:
>     * Is this also needed for wasi-threads?

I don't think so. I was mostly thinking about browsers for this change, but I see no reason why it also wouldn't work for wasi-threads. I realize that I disabled the `wait()` family of functions for now. Once WASM atomics are stable I'll re-enable them.

Aside from that, I've fixed the issues you brought up.

taiki-e (approved):

PR #81, opened 2023-09-10
Disable IntoNotification::tag with no_std

taiki-e:
https://github.com/smol-rs/event-listener/pull/75 disabled only IntoNotification::tag_with, but IIUC, this doesn't work either, right?


notgull (approved):
Yes, you're right, thanks!

Since `Tag` can't really be accessed by the user, it's not a *huge* deal, but it's still cleaner to disable it.

PR #83, opened 2023-09-10
Relax memory ordering in AtomicCell::replace

taiki-e:
Assuming we are not depending on the properties of SeqCst here, this should be sufficient.


notgull (approved):
Thanks!

Issue #84, opened 2023-10-05
Miri test failure with --no-default-features


taiki-e:
https://github.com/smol-rs/event-listener/actions/runs/6367362187/job/17285871726

```
failures:

---- sys::tests::listener_slab_notify_prop stdout ----
thread 'sys::tests::listener_slab_notify_prop' panicked at src/no_std.rs:1270:9:
assertion `left == right` failed
  left: Listener { state: Task(_), prev: Cell { value: None }, next: Cell { value: Some(3) } }
 right: Listener { state: Task(_), prev: Cell { value: None }, next: Cell { value: Some(3) } }
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:72:14
   2: core::panicking::assert_failed_inner
             at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:[31](https://github.com/smol-rs/event-listener/actions/runs/6367362187/job/17285871726#step:6:32)5:17
   3: core::panicking::assert_failed::<sys::Entry<()>, sys::Entry<()>>
             at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/panicking.rs:270:5
   4: sys::tests::listener_slab_notify_prop
             at src/no_std.rs:1270:9
   5: sys::tests::listener_slab_notify_prop::{closure#0}
             at src/no_std.rs:1119:[36](https://github.com/smol-rs/event-listener/actions/runs/6367362187/job/17285871726#step:6:37)
   6: <{closure@src/no_std.rs:1119:5: 1361:6} as std::ops::FnOnce<()>>::call_once - shim
             at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
   7: <fn() -> std::result::Result<(), std::string::String> as std::ops::FnOnce<()>>::call_once - shim(fn() -> std::result::Result<(), std::string::String>)
[...]

PR #86, opened 2023-10-12
Update fmt::Debug to produce new info

### LINKED ISSUE #79: fmt::Debug should produce actually useful output
At the moment, the `fmt::Debug` implementation for `Event` and `EventListener` just writes `Event { .. }`, giving no useful information. It would be better if it produced actual descriptive information regarding the contents of the `Event`.

For `Event`, if it can lock the mutex, I'd like it to produce:

- Number of notified listeners.
- Total number of listeners.

For `EventListener`, it should indicate whether it's registered in the `Event` and, if it is, its current state (listening, notified, etc etc).

scirin:
Closes  #79 
 Let me know what to change

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -181,7 +181,7 @@ impl<T> fmt::Debug for Event<T> {
                     .field("listeners_total", &total_count)
                     .finish()
             }
-            None => f.debug_tuple("event").field(&"<uninitialized>").finish(),
+            None => f.debug_tuple("event").field(&format_args!("<uninitialized>")).finish(),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,14 +116,14 @@ struct Inner<T> {
     /// The number of notified entries, or `usize::MAX` if all of them have been notified.
     ///
     /// If there are no entries, this value is set to `usize::MAX`.
-    notified: AtomicUsize,
+    pub notified: AtomicUsize,
 
     /// Inner queue of event listeners.
     ///
     /// On `std` platforms, this is an intrusive linked list. On `no_std` platforms, this is a
     /// more traditional `Vec` of listeners, with an atomic queue used as a backup for high
     /// contention.
-    list: sys::List<T>,
+    pub list: sys::List<T>,
 }
 
 impl<T> Inner<T> {
@@ -161,7 +161,6 @@ pub struct Event<T = ()> {
     /// is an `Arc<Inner>` so it's important to keep in mind that it contributes to the [`Arc`]'s
     /// reference count.
     inner: AtomicPtr<Inner<T>>,
-    listener_count: AtomicUsize,
 }
 
 unsafe impl<T: Send> Send for Event<T> {}
@@ -172,19 +171,16 @@ impl<T> core::panic::RefUnwindSafe for Event<T> {}
 
 impl<T> fmt::Debug for Event<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let mut ds = f.debug_struct("Event");
+        let inner = self.inner();
+        let notified_count = unsafe { (*inner).notified.load(Ordering::Relaxed) };
+        let total_count = unsafe { (*inner).list.total_listeners() };
 
-        if let Some(inner) = self.try_inner() {
-            let notified_listeners = inner.notified.load(Ordering::Acquire);
-            let total_listeners = self.listener_count.load(Ordering::Relaxed);
-
-            ds.field("Notified Listeners", &notified_listeners);
-            ds.field("Total Listeners count", &total_listeners);
-        } else {
-            ds.field("Status", &"Not Initialized");
-        }
-
-        ds.finish()
+        write!(
+            f,
+            "Event {{\n  Number of notified listeners: {}\n  Total number of listeners: {}\n}}",
+            notified_count,
+            total_count
+        )
     }
 }
 
@@ -213,16 +209,7 @@ impl<T> Event<T> {
     pub const fn with_tag() -> Self {
         Self {
             inner: AtomicPtr::new(ptr::null_mut()),
-            listener_count: AtomicUsize::new(0),
         }
-    }
-
-    pub fn add_listener(&self) {
-        self.listener_count.fetch_add(1, Ordering::Relaxed);
-    }
-
-    pub fn remove_listener(&self) {
-        self.listener_count.fetch_sub(1, Ordering::Relaxed);
     }
     /// Tell whether any listeners are currently notified.
     ///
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -116,14 +116,14 @@ struct Inner<T> {
...
-    notified: AtomicUsize,
+    pub notified: AtomicUsize,
 
     /// Inner queue of event listeners.
     ///
     /// On `std` platforms, this is an intrusive linked list. On `no_std` platforms, this is a
     /// more traditional `Vec` of listeners, with an atomic queue used as a backup for high
     /// contention.
-    list: sys::List<T>,
+    pub list: sys::List<T>,
```
notgull, line 126: These fields don't need to be `pub`, I think.

src/lib.rs
```diff
@@ -171,7 +171,15 @@ impl<T> core::panic::RefUnwindSafe for Event<T> {}
 
 impl<T> fmt::Debug for Event<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str("Event { .. }")
+        let inner = self.inner();
```
notgull, line 174: Instead of calling `inner()`, call `try_inner()`, and if it returns `None` output `Event(<uninitialized>)` using `debug_tuple`. See [here](https://github.com/smol-rs/async-executor/blob/2cfb6e4ed03c7eaf6c5fb2d3170ac870ba0a9b9d/src/lib.rs#L981-L995) for an example.

src/lib.rs
```diff
@@ -650,7 +656,15 @@ unsafe impl<T: Send> Sync for EventListener<T> {}
...
 impl<T> fmt::Debug for EventListener<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str("EventListener { .. }")
+        let mut ds = f.debug_struct("EventListener");
+
+        if self.is_listening() {
+            ds.field("Listening", &"Yes");
+        } else {
+            ds.field("Listening", &"No");
+        }
```
notgull, line 665: Instead of `yes` and `no`, I would just use `true` and `false`, which can be done by just using `self.is_listening()` as the field here.

src/lib.rs
```diff
@@ -202,7 +210,6 @@ impl<T> Event<T> {
             inner: AtomicPtr::new(ptr::null_mut()),
         }
     }
-
```
notgull, line 205: ?

src/lib.rs
```diff
@@ -456,7 +463,6 @@ impl Event<()> {
...
```
notgull, line 459: ?

src/no_std.rs
```diff
@@ -240,6 +240,14 @@ impl<T> List<T> {
...
         }
     }
+    pub fn total_listeners(&self) -> usize {
+        self.inner
+            .try_lock()
+            .as_ref()
+            .map(|lock| &**lock)
+            .map(|listener_slab| listener_slab.listeners.len())
+            .unwrap_or(0)
+    }
```
notgull, line 250: This returns something different from the `total_listeners` in `std.rs`.

src/no_std.rs
```diff
@@ -240,6 +240,14 @@ impl<T> List<T> {
             queue: concurrent_queue::ConcurrentQueue::unbounded(),
         }
     }
+    pub fn total_listeners(&self) -> usize {
+        self.inner
+            .try_lock()
+            .as_ref()
+            .map(|lock| &**lock)
+            .map(|listener_slab| listener_slab.listeners.len())
```
notgull, line 248: Don't use `map()` twice here, just get `listeners` from `lock` itself.

src/lib.rs
```diff
@@ -171,7 +171,15 @@ impl<T> core::panic::RefUnwindSafe for Event<T> {}
...
-        f.write_str("Event { .. }")
+        let inner = self.inner();
+        let notified_count = unsafe { (*inner).notified.load(Ordering::Relaxed) };
+        let total_count = unsafe { (*inner).list.total_listeners() };
+
+        write!(
+            f,
+            "Event {{\n  Number of notified listeners: {:?}\n  Total number of listeners: {:?}\n}}",
+            notified_count, total_count
+        )
```
notgull, line 182: To simplify, use `debug_struct` to construct the output instead of manually formatting like this.

notgull (changes requested):
I definitely like `Event` including the number of notified listeners and the number of total listeners. However, the debug output fields should use `snake_case`, e.g. `notified_listeners: 7`.

For `EventListener`, `State` indicates the states that the listener can be in. I think that this state should be reflected in the debug output.

notgull:
Rebase on the current `master` to fix the current CI error (#84)

scirin:
Let me know if I missed something

src/lib.rs
```diff
@@ -171,7 +171,18 @@ impl<T> core::panic::RefUnwindSafe for Event<T> {}
...
+            Some(inner) => {
+                let notified_count = inner.notified.load(Ordering::Relaxed);
+                let total_count = inner.list.total_listeners();
+
+                f.debug_struct("Event")
+                    .field("listeners_notified", &notified_count)
+                    .field("listeners_total", &total_count)
+                    .finish()
+            }
+            None => f.debug_tuple("event").field(&"<uninitialized>").finish(),
```
notgull, line 184:
```suggestion
            None => f.debug_tuple("event").field(&format_args!("<uninitialized>")).finish(),
```

This makes it so the `<uninitialized>` text isn't in quotes.

src/lib.rs
notgull: Can you add back these lines to reduce diff noise?

src/std.rs
```diff
@@ -44,6 +44,10 @@ impl<T> List<T> {
             notified: 0,
         }))
     }
+    // Accessor method because fields are private, not sure how to go around it
+    pub fn total_listeners(&self) -> Result<usize, &str> {
+        self.0.lock().map(|mutex| mutex.len).map_err(|_| "<locked>")
```
notgull, line 49: Can you do `try_lock()` to avoid contention?

notgull (changes requested):

src/lib.rs
```diff
@@ -171,7 +171,21 @@ impl<T> core::panic::RefUnwindSafe for Event<T> {}
 
 impl<T> fmt::Debug for Event<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str("Event { .. }")
+        match self.try_inner() {
+            Some(inner) => {
+                let notified_count = inner.notified.load(Ordering::Relaxed);
+                let total_count = inner.list.total_listeners();
```
notgull, line 177:
```suggestion
                let total_count = match inner.list.total_listeners() {
                    Ok(total_count) => total_count,
                    Err(_) => return f.debug_tuple("Event").field(&format_args!("<locked>")).finish(),
                };
```

notgull (changes requested):
Almost good, just a few nitpicks.

scirin:
fixed it all

notgull (approved):
Thanks!

PR #85, opened 2023-10-08
Migrate to Rust 2021

taiki-e:
Our MSRV (1.59) is higher than 1.56 that Rust 2021 was stabilized.

PR #88, opened 2023-10-16
v3.0.1

notgull:
- Migrate to Rust 2021 (#85)


taiki-e:
#85 has no effect on the public API, so I think there is no need to create a new release just for that.

PR #90, opened 2023-10-18
Document that EventListeners must be listen'd on

### LINKED ISSUE #89: EventListener::new footgun
Copy&pasting from the matrix channel to avoid repeating:

> zeenix: this method takes an Event ref but doesn't actually setup the listener to listen to this event. This seems like a pretty big footgun
> Saltbag: The only other option is to have it insert itself into the list on await, which has its own footguns. It can’t be set up in new since it needs to be pinned.
> zeenix: it's just a very strange and unintuitive api now. I see multiple issues here:
> 
> * `EventListener` existing w/o listening is just strange. The name implies something that's listening.
> * `new` takes a `Event` ref but doesn't listen to the given event.
> * The footgun of awaiting a listener w/o calling listen on it. That state should not be allowed to ever exist.
> 
> If listening require pinning, that should always be implied then. That's the whole point of `EventListener` type. At the very least, new and `EventListener` docs should document all these caveats well.

notgull:
In retrospect it is sometimes unclear in the documentation that the new event listener needs to be pinned and inserted into the list before it can receive events. This PR adds documentation that should clarify this issue.

Closes #89
cc @zeenix

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -658,10 +658,10 @@ pin_project_lite::pin_project! {
     /// another active listener. Whether one *additional* listener will be notified depends on what
     /// kind of notification was delivered.
     ///
-    /// The listener is not registered into the linked list inside of the [`Event`] by default. It
-    /// needs to be pinned first before being inserted using the `listen()` method. After the
-    /// listener has begun `listen`ing, the user can `await` it like a future or call `wait()`
-    /// to block the current thread until it is notified.
+    /// The listener is not registered into the linked list inside of the [`Event`] by default if
+    /// it is created via the `new()` method. It needs to be pinned first before being inserted
+    /// using the `listen()` method. After the listener has begun `listen`ing, the user can
+    /// `await` it like a future or call `wait()` to block the current thread until it is notified.
     ///
     /// ## Examples
     ///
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -638,6 +657,77 @@ pin_project_lite::pin_project! {
     /// If a notified listener is dropped without receiving a notification, dropping will notify
     /// another active listener. Whether one *additional* listener will be notified depends on what
     /// kind of notification was delivered.
+    ///
+    /// The listener is not registered into the linked list inside of the [`Event`] by default. It
```
zeenix, line 661: Nit: we should make it explicit here that we're talking about EventListener created using `new` method and not `Event::listen`. "by default" is a bit unclear.

zeenix (changes requested):
Thanks for the quick action here and adding such detailed docs! :+1: However, apart from the nit comment, I think we should also at least change the API (in a compatible way) to either add a new `new` method that doesn't take a `Event` ref and deprecate `new` or (if possible) make `listen()` implicit in the first time `EventListener` is polled and deprecate `EventListener::listen`. I'd very much prefer the latter, since it removes the footgun.

zeenix:
> think we should also at least change the API (in a compatible way) to either add a new `new` method that doesn't take a `Event` ref and deprecate `new` or (if possible) make `listen()` implicit in the first time `EventListener` is polled and deprecate `EventListener::listen`. I'd very much prefer the latter, since it removes the footgun.

I created #91 for this part so you can ignore this comment here.

zeenix (approved):
LGTM, thanks again.

Issue #91, opened 2023-10-18
Remove EventListener::new footgun


zeenix:
As discussed in #90 and #89, `EventListener::new` has a big footgun: it takes `&Event` ref but doesn't add itself to listen to the event passed. This is highly unintuitive IMO and if possible, we should remedy this. Two options here:

1. Add a new method that doesn't take a Event ref and deprecate `new` in favor of that. OR
2. (if possible) make `listen()` implicit in the first poll of `EventListener` and deprecate `EventListener::listen`.

I'd very much prefer the 2, since it removes the footgun completely.

notgull:
I agree with the idea that `new()` should take no parameters and `listen()` should take the `&Event` parameter; it makes more sense that way. Unfortunately that would be breaking change as this point, and deprecation/introduction of new methods introduces its own footguns, as mixing the two types of methods would lead to unexpected panics. However, I disagree with the second option, as it would make polling/waiting on the `EventListener` much more tricky.

If the listener is linked into the list on the first poll of the `Future`, it makes it more difficult to receive events. Until the listener is fully linked into the list, it will not receive events. Meaning, if an event occurs *while* the listener is being atomically linked into the list, it will not receive that event. This makes it necessary to check any atomic flags *before* and *after* the list is inserted. This could would become necessary:

```rust
loop {
    if check_the_flag() {
        break;
    }

    let mut listener = EventListener::new(&event);
    pin!(listener);

    // Poll the future once to insert it into the list.
    future::poll_once(listener.as_mut()).await;

    if check_the_flags() {
        break;
    }

    // Now we can fully await on the listener.
    listener.await;
}
```

This introduces a much larger footgun, however, as a naive user may just try to `.await` on the future without making sure the condition that they're waiting on hasn't changed after the insert. This code *will* deadlock unpredictably:

```rust
loop {
    if check_the_flag() {
        break;
    }

    let listener = EventListener::new(&event);
    pin!(listener);
    listener.await;
}
```

Inserting on first poll also neglects the `wait()`, use case, which doesn't really have a "poll" mechanism at all. So the above deadlock would become the mandatory way of using `wait()`, unless you do `wait_timeout(Duration::from_secs(0))` or something to "register" it first. But that adds its own additional complexity.

[...]

PR #92, opened 2023-10-29
v3.0.1

notgull:
- Emphasize that `listen()` must be called on `EventListener` in documentation. (#90)
- Write useful output in `fmt::Debug` implementations. (#86)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Version 3.0.1
+
+- Emphasize that `listen()` must be called on `EventListener` in documentation. (#90)
+- Write useful output in `fmt::Debug` implementations. (#86)
+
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -782,17 +782,15 @@ impl<T> EventListener<T> {
     /// use event_listener::{Event, EventListener};
     ///
     /// let event = Event::new();
-    /// let listener = EventListener::new(&event);
```
taiki-e, line 785: This seems to be a breaking change.

notgull: Whoops, I accidentally pushed from a branch where I was experimenting with https://github.com/smol-rs/event-listener/issues/91.

taiki-e (approved):

PR #94, opened 2023-11-06
Fix the EventListener::new() footgun

### LINKED ISSUE #91: Remove EventListener::new footgun

notgull:
This is a breaking change. It makes `new()` take no parameters in its
signature and `listen()` take a reference to an event in its signature.
This should avoid a footgun where a listener can be waited on without
listening on it.

Closes #91
cc @zeenix

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -9,7 +9,7 @@ const COUNT: usize = 8000;
 fn bench_events(c: &mut Criterion) {
     c.bench_function("notify_and_wait", |b| {
         let ev = Event::new();
-        let mut handles = iter::repeat_with(|| EventListener::new())
+        let mut handles = iter::repeat_with(EventListener::new)
             .take(COUNT)
             .collect::<Vec<_>>();
 
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -9,7 +9,7 @@ const COUNT: usize = 8000;
 fn bench_events(c: &mut Criterion) {
     c.bench_function("notify_and_wait", |b| {
         let ev = Event::new();
-        let mut handles = iter::repeat_with(|| EventListener::new(&ev))
+        let mut handles = iter::repeat_with(|| EventListener::new())
             .take(COUNT)
             .collect::<Vec<_>>();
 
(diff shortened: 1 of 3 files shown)


notgull:
This causes a slightly performance regression.

```
notify_and_wait         time:   [549.58 µs 549.97 µs 550.42 µs]
                        change: [+76.620% +77.217% +77.648%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 16 outliers among 100 measurements (16.00%)
  6 (6.00%) low severe
  7 (7.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
```

zeenix:
> This causes a worrying performance regression.

* Is this always reproducible? I found criterian-based benchmarking to be a bit unreliable.
* Where is the bottleneck?

notgull:
> * Where is the bottleneck?

The bottleneck comes from the additional `Option` around the `Listener` type in `EventListener`. Let me try shuffling it around and seeing if the compiler likes it better.

notgull:
```
$ cargo bench                 
   Compiling event-listener v3.0.1 (/home/jtnunley/Projects/smol-rs/event-listener)
    Finished bench [optimized] target(s) in 3.15s
     Running benches/bench.rs (/hard_data/cargo_target/release/deps/bench-eeda2fe5c3372937)
notify_and_wait         time:   [207.45 µs 207.70 µs 207.95 µs]
                        change: [+7.4551% +7.6268% +7.8167%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

I find this regression much more palatable.

zeenix (approved):
Thanks for doing this work and making sure there is no significant performance hit. :+1:

PR #96, opened 2023-11-12
Implement `UnwindSafe` for `EventListener`

### LINKED ISSUE #95: `EventListener` doesn't implement `UnwindSafe`
I think it's a pretty normal use case to create an `EventListener`, then send the `EventListener` somewhere else, but this somewhere else panics and we want to continue running the program anyway.

Since `Event` itself implements `RefUnwindSafe`, it seems that `EventListener` not implementing `UnwindSafe` is just an overlook?

tomaka:
Close #95 

Following to the comment in #95 I'm opening this very naive PR. As I'm not familiar with the internals and that it's non-trivial, I don't actually know if this is correct.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -761,6 +761,9 @@ pin_project_lite::pin_project! {
 unsafe impl<T: Send> Send for EventListener<T> {}
 unsafe impl<T: Send> Sync for EventListener<T> {}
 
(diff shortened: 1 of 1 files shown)


notgull (approved):
Thanks!

I think this is correct. v2 implemented these traits for the `EventListener`, so not implementing them here was an oversight on my part.

PR #98, opened 2023-11-18
v4.0.0

notgull:
- **Breaking:** Fix a footgun in the `EventListener` type. `EventListener::new()` now no longer takes an `&Event` as an argument, and `EventListener::listen()` takes the  `&Event` as an argument. Hopefully this should prevent `.await`ing on a listener without making sure it's listening first. (#94)


zeenix (approved):
LGTM. Thanks!

PR #103, opened 2023-12-19
v4.0.1

notgull:
- Fix a use-after-move error after an `EventListener` is assigned to listen to another `Event`. (#101)


zeenix (approved):
Great, thanks!

PR #105, opened 2023-12-20
Setup a faster, less footgun-prone API

### LINKED ISSUE #104: RFC: Less complex, footgun free API
Continued from the discussion in #100 

The APIs of v3.x and v4.x increase complexity at the cost of adding more footguns. Originally I'd wanted to keep the API small and simple, but this appears to have introduced a handful of different ways to mess up (e.g. panicking when creating a `new` `EventListener` without `listen`ing on it). In addition, v4.0 (#94) introduced a not-insignificant amount of overhead over the v3.x API. Therefore, as much as it pains me to have three breaking changes in this crate in such a short span of time, I think we need a better API.

*Note: I am bad at naming things. None of these names are final.*

## Reference-Level Explanation

The idea I proposed in #100: is as follows: add an API based around a `Listener` trait that would look like this:

```rust
pub trait Listener<T>: Future<Output = T> + Sealed {
    fn wait(self) -> T;
    fn wait_timeout(self, timeout: Duration) -> Option<T>;
    fn wait_deadline(self, deadline: Instant) -> Option<T>;

    fn discard(self) -> Option<T>;
    fn listens_to(&self, event: &Event<T>) -> bool;
    fn same_event(&self, other: &(impl Listener<T> + ?Sized)) -> bool;
}
```

This `Listener` trait can do anything that the current `EventListener` type can do. It can be `await`ed or it can be used to block the current thread. Alternatively, it can be discarded.

To emulate the previous `2.x` API, we would have an `EventListener` type. This is a heap-allocated event listener that can be moved freely.

```rust
impl<T> Event<T> {
    pub fn listen(&self) -> EventListener<T>;
}

pub struct EventListener<T = ()>(/* ... */);

impl Listener<T> for EventListener<T> { /* ... */ }
```

It would be used like this:

```rust
struct Flag {
    flag: AtomicBool,
    change_ops: Event<()>
}

impl Flag {
    async fn wait(&self) {
        loop {
            // Check the flag.
            if self.flag.load(Ordering::Relaxed) {
                return;
            }

            // Create a listener. By default it is already inserted into the list.
            let listener = self.change_ops.listen();

            // Check the flag again.
            if self.flag.load(Ordering::SeqCst) {
                return;
            }

            // Wait on the listener.
            listener.await;
        }
    }

    fn notify(&self) {
        self.flag.store(true, Ordering::Release);
        self.change_ops.notify(1);
    }
}
```

However, this is inefficient, as it preforms a heap allocation every time `listen()` is called. We also provide an API that allows one to use the stack instead of the heap, at a slightly higher complexity cost. To start, you create a `StackSlot`, which contains all of the state of the `EventListener` but stored on the stack. After being pinned, it can be transformed into a `StackListener`.

```rust
impl<T> Event<T> {
    pub fn stack_slot(&self) -> StackSlot<'_, T> { /* ... */ }
}

pub struct StackSlot<'ev, T>(/* ... */);
pub struct StackListener<'ev, 'listener, T>(/* ... */);

impl<'ev, T> StackSlot<'ev, T> {
    pub fn listen<'this>(self: Pin<&'this mut Self>) -> StackListener<'ev, 'this, T> { /* ... */ }
}

impl<T> Listener<T> for StackListener<'_, '_, T> { /* ... */ }
```

In addition, a macro will be provided that creates a `StackSlot` from an `Event` and automatically pins it to the heap.

```rust
macro_rules! listener {
    ($name:ident, $event:expr) => { /* ... */ }
}

/*
listener!(l, ev);
Expands to:
let mut l = ev.stack_slot();
let mut l = Pin::new_unchecked(&mut l);
*/
```

The example from above can be modified to look like this:

```rust
struct Flag {
    flag: AtomicBool,
    change_ops: Event<()>
}

impl Flag {
    async fn wait(&self) {
        listener!(slot, &self.change_ops);

        loop {
            // Check the flag.
            if self.flag.load(Ordering::Relaxed) {
                return;
            }

            // Insert the listener into the list.
            let listener = slot.listen();

            // Check the flag again.
            if self.flag.load(Ordering::SeqCst) {
                return;
            }

            // Wait on the listener.
            listener.await;
        }
    }

    fn notify(&self) {
        self.flag.store(true, Ordering::Release);
        self.change_ops.notify(1);
    }
}
```

@smol-rs/admins Thoughts on this?

notgull:
This PR sets up a new API that aims to implement #104. It sets up a Listener trait, which is implemented for a heap-based EventListener and a stack-based StackListener. The latter is only able to be instantiated through the "listener!" macro.

While writing this, I realized that this breaks the hand-rolled futures that `async-channel` and `async-lock` use nowadays. For now we can burn that bridge when we get there and just use `EventListener` in the hand-rolled futures.

Closes #104

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -1,7 +1,7 @@
 use std::iter;
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use event_listener::{prelude::*, Event};
+use event_listener::{Event, Listener};
 
 const COUNT: usize = 8000;
 
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -1,7 +1,7 @@
 use std::iter;
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use event_listener::Event;
+use event_listener::{prelude::*, Event};
 
 const COUNT: usize = 8000;
 
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -1,31 +1,22 @@
 use std::iter;
-use std::pin::Pin;
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use event_listener::{Event, EventListener};
+use event_listener::Event;
 
 const COUNT: usize = 8000;
 
 fn bench_events(c: &mut Criterion) {
     c.bench_function("notify_and_wait", |b| {
         let ev = Event::new();
-        let mut handles = iter::repeat_with(EventListener::new)
-            .take(COUNT)
-            .collect::<Vec<_>>();
+        let mut handles = Vec::with_capacity(COUNT);
 
         b.iter(|| {
-            for handle in &mut handles {
-                // SAFETY: The handle is not moved out.
-                let listener = unsafe { Pin::new_unchecked(handle) };
-                listener.listen(&ev);
-            }
+            handles.extend(iter::repeat_with(|| ev.listen()).take(COUNT));
 
             ev.notify(COUNT);
 
-            for handle in &mut handles {
-                // SAFETY: The handle is not moved out.
-                let listener = unsafe { Pin::new_unchecked(handle) };
-                listener.wait();
+            for handle in handles.drain(..) {
+                handle.wait();
             }
         });
     });
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,7 @@
 //! use std::thread;
 //! use std::time::Duration;
 //! use std::usize;
-//! use event_listener::Event;
+//! use event_listener::{Event, prelude::*};
 //!
 //! let flag = Arc::new(AtomicBool::new(false));
 //! let event = Arc::new(Event::new());
@@ -108,7 +108,7 @@ pub use notify::{IntoNotification, Notification};
 
 /// Useful traits for notifications.
 pub mod prelude {
-    pub use crate::{IntoNotification, Notification};
+    pub use crate::{IntoNotification, Listener, Notification};
 }
 
 /// Inner state of [`Event`].
@@ -213,7 +213,7 @@ impl<T> Event<T> {
     /// # Examples
     ///
     /// ```
-    /// use event_listener::Event;
+    /// use event_listener::{Event, prelude::*};
     ///
     /// let event = Event::<usize>::with_tag();
     /// ```
@@ -230,7 +230,7 @@ impl<T> Event<T> {
     /// # Examples
     ///
     /// ```
-    /// use event_listener::Event;
+    /// use event_listener::{Event, prelude::*};
     ///
     /// let event = Event::new();
     /// let listener = event.listen();
@@ -254,7 +254,7 @@ impl<T> Event<T> {
     /// # Examples
     ///
     /// ```
-    /// use event_listener::Event;
+    /// use event_listener::{Event, prelude::*};
     ///
     /// let event = Event::new();
     /// let listener = event.listen();
@@ -283,7 +283,7 @@ impl<T> Event<T> {
         let inner = ManuallyDrop::new(unsafe { Arc::from_raw(self.inner()) });
 
         // Allocate the listener on the heap and insert it.
-        let mut listener = Box::pin(Listener {
+        let mut listener = Box::pin(InnerListener {
             event: Arc::clone(&inner),
             listener: None,
         });
@@ -322,7 +322,7 @@ impl<T> Event<T> {
     /// Use the default notification strategy:
     ///
     /// ```
-    /// use event_listener::Event;
+    /// use event_listener::{Event, prelude::*};
     ///
     /// let event = Event::new();
     ///
@@ -539,7 +539,7 @@ impl Event<()> {
     /// # Examples
     ///
     /// ```
-    /// use event_listener::Event;
+    /// use event_listener::{Event, prelude::*};
     ///
     /// let event = Event::new();
     /// ```
@@ -576,7 +576,7 @@ impl Event<()> {
     /// # Examples
     ///
     /// ```
-    /// use event_listener::Event;
+    /// use event_listener::{Event, prelude::*};
     /// use std::sync::atomic::{self, Ordering};
     ///
     /// let event = Event::new();
@@ -628,7 +628,7 @@ impl Event<()> {
     /// # Examples
     ///
     /// ```
-    /// use event_listener::Event;
+    /// use event_listener::{Event, prelude::*};
     ///
     /// let event = Event::new();
     ///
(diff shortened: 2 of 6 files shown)


benches/bench.rs
```diff
@@ -1,7 +1,7 @@
 use std::iter;
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use event_listener::Event;
+use event_listener::{prelude::*, Event};
```
zeenix, line 4: I always find preludes annoying (except in case of std) since their use makes it hard to read the code (especially if there are multiple preludes being imported). I know you introduced them in e001c7a0 so it's not super relevant to this PR but I would like to know if you feel very strongly in favor of them. If not, I'd like to remove them.

src/lib.rs
```diff
@@ -866,6 +886,18 @@ impl<T> Future for EventListener<T> {
     }
 }
 
+/// Create a stack-based event listener for an [`Event`].
```
zeenix, line 891: A simple example would be super.

zeenix (changes requested):
Really nice work. I believe we're now getting to an API that's simple, sound and at the same time allows to be easily heap-free. :+1: 

Some nitpicks though. I do realize this is still a draft but thought you'd appreciate early feedback. :)

benches/bench.rs
notgull: I added them mostly to make importing the at-the-time new `IntoNotification` trait easier. If you think they should be removed I'm fine with that.

zeenix: > I added them mostly to make importing the at-the-time new `IntoNotification` trait easier.

Yeah, I understand. They do make it easier to write but there is a cost to readability. Modern editors make it easier for your in most cases to import the needed traits anyway.

> If you think they should be removed I'm fine with that.

Yes, please.

notgull:
It appears that this somehow introduces a deadlock into `async-lock`. See the test failure here: https://github.com/smol-rs/async-lock/actions/runs/7493609820/job/20509520336

zeenix:
> It appears that this somehow introduces a deadlock into `async-lock`. See the test failure here: https://github.com/smol-rs/async-lock/actions/runs/7493609820/job/20509520336

Any progress on this? If this is proving difficult and will take time, perhaps we can first have other crates ported to event-handler 4 so folks don't end up depending on multiple versions of event-handler?

notgull:
> Any progress on this?

None yet. At the moment I'm still trying to replicate the deadlock using only `event-listener`.

> If this is proving difficult and will take time, perhaps we can first have other crates ported to event-handler 4 so folks don't end up depending on multiple versions of event-handler?

I would agree with this. Most of the smol crates are already ported to event-listener v4.

zeenix:
> > If this is proving difficult and will take time, perhaps we can first have other crates ported to event-handler 4 so folks don't end up depending on multiple versions of event-handler?
> 
> I would agree with this. Most of the smol crates are already ported to event-listener v4.

Cool. I'll see if I can provide a PR for async-broadcast. Any others left?

notgull:
Not to my knowledge

notgull:
At the moment I think the deadlock is a Miri bug, so I've disabled those tests in https://github.com/smol-rs/async-lock/pull/74/commits/1d97f8a8879e069f2438627cad64f7354d5db1e1. I think this should be ready now.

src/lib.rs
```diff
@@ -934,19 +747,16 @@ impl<T> EventListener<T> {
     ///
     /// ```
     /// use std::time::Duration;
-    /// use event_listener::Event;
+    /// use event_listener::{Event, IntoNotification, Listener};
```
zeenix, line 750: * Are these imports really needed here? I don't see a use of `Listener` or calls to `IntoListener` trait. :thinking: 
* We didn't discuss `IntoNotification` in #104 btw. I'm guessing the API we agreed on there, didn't work out exactly as planned?

zeenix (changes requested):

src/lib.rs
notgull: > * Are these imports really needed here? I don't see a use of `Listener` or calls to `IntoListener` trait. 🤔

Good catch, fixed

>     * We didn't discuss `IntoNotification` in [RFC: Less complex, footgun free API #104](https://github.com/smol-rs/event-listener/issues/104) btw. I'm guessing the API we agreed on there, didn't work out exactly as planned?

I'm still a fan of the current `IntoNotification` API, so I think we should keep it.

zeenix: > I'm still a fan of the current `IntoNotification` API, so I think we should keep it.

I think it's more complex than the design you proposed. With the proposed design in #104, `Listener` just becomes a trait that 2 types implement and that's it. Unless you feel very strongly about this (you're doing the work so your strong preferences trumps others :laughing:), I'd say we ask in #104 of what others think of your modified proposal.

notgull: I'm not clear what is "modified" here? `Listener` was mentioned in #104.

zeenix: `Listener` was, yes. Not `IntoListener`. The latter implies a conversion to a listener, while the former does not and hence why I think it's a simpler concept/API.

src/notify.rs
```diff
@@ -342,7 +342,7 @@ impl<T, F: FnMut() -> T> TagProducer for F {
 /// into this:
 ///
 /// ```
-/// use event_listener::{Event, prelude::*};
+/// use event_listener::{Event, IntoNotification, Listener};
```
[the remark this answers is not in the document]
zeenix: This specific one here is still there. Maybe I'm missing something and this one is needed? :thinking:

src/lib.rs
zeenix: Having said that, if typical user doesn't have to use/import `IntoListener`, I'm not too strongly against it.

notgull: Ah, `IntoNotification`. I think that's a separate topic entirely from this API. There are alternatives for it in the API; it's just there to allow for easily composable tagged events.

src/notify.rs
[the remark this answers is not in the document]
zeenix: still there. :wink: Let me know if you need help with removing all unneeded `IntoListener` imports.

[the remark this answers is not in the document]
notgull: It is actually needed for this example.

https://github.com/smol-rs/event-listener/blob/b1a2273d3653e607c2ff3aae237b86413801a725/src/notify.rs#L350

src/lib.rs
```diff
@@ -230,7 +225,7 @@ impl<T> Event<T> {
     /// # Examples
     ///
     /// ```
-    /// use event_listener::Event;
+    /// use event_listener::{Event, Listener};
```
zeenix, line 228: `Listener` needed here?

zeenix (approved):
alright, let's do it. :)

PR #107, opened 2023-12-25
Avoid spinning when waiting for deadline

AntonSol919:
wait_deadline uses 100% CPU. 

I'm unfamiliar with the intricacies as this fix requires `parking #[cfg(not(loom))]`. 
If spinning was done intentionally it should probably be documented.


notgull (approved):
Thanks!

PR #112, opened 2024-01-07
Remove RefCell<Option> from PARKER

taiki-e:


taiki-e:
Nah, it is better to initialize the parker in the TLS initialization code than to do this.

taiki-e:
Ok, removed both RefCell and Option from it.

taiki-e:
https://github.com/smol-rs/event-listener/blob/b1d437ae83eec859768b6170ea68619c5b2bb6e9/src/lib.rs#L1060-L1062

Is RefCell used to detect reentrant parking? This seems true in other places where we need to wait for arbitrary futures, but that should not happen in event-listener, right?

- https://github.com/smol-rs/futures-lite/blob/2e555681d54dedb5f18455deb8d403386a368729/src/future.rs#L89
- https://github.com/smol-rs/async-io/blob/1c635937427c70300118bd8718b4c6c5c8645da8/src/driver.rs#L186

notgull (approved):
Thanks!

notgull:
> Is RefCell used to detect reentrant parking? This seems true in other places (futures-lite, async-io) where we need to wait for arbitrary futures, but that should not happen in event-listener, right?


This is right, since you can't run arbitrary code while parked.

Issue #106, opened 2023-12-23
expose listener count


beckend:
Using this as a base to create libraries, it's very useful to know the listener count from time to time.

notgull:
This can be done with relative ease on `std` platforms, but on `no_std` it's impossible to know how many listeners are in the last at any given time.

beckend:
Sure, then a feature flag?

notgull:
It sounds fine to restrict it to `std` only. I'll accept a PR

beckend:
PR is online

PR #114, opened 2024-01-16
add method total_listeners() when feature std

beckend:
https://github.com/smol-rs/event-listener/issues/106

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -502,15 +502,14 @@ impl<T> Event<T> {
     /// drop(listener1);
     /// drop(listener2);
     /// assert_eq!(event.total_listeners().unwrap(), 0);        
-    ///
     /// ```
     #[cfg(feature = "std")]
     #[inline]
-    pub fn total_listeners(&self) -> Result<usize, String> {
+    pub fn total_listeners(&self) -> usize {
         if let Some(inner) = self.try_inner() {
             inner.list.total_listeners_wait()
         } else {
-            Ok(0)
+            0
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -480,8 +480,10 @@ impl<T> Event<T> {
         inner
     }
 
-    /// Return the listener count by acquiring a lock
-    /// This is just a snapshot of the number of listeners at this point in time. It is possible for the actual number to change at any point.
+    /// Return the listener count by acquiring a lock.
+    ///
+    /// This is just a snapshot of the number of listeners at this point in time.
+    /// It is possible for the actual number to change at any point.
     /// The number should only ever be used as a hint.
     /// This is only available when `std` feature is enabled.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -480,7 +480,7 @@ impl<T> Event<T> {
         inner
     }
 
-    /// Return the listener count by aquiring a lock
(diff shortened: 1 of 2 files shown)


src/std.rs
```diff
@@ -54,6 +54,14 @@ impl<T> List<T> {
             Err(_) => Err("<locked>"),
         }
     }
+
+    // Get the listener count by blocking
+    pub(crate) fn total_listeners_wait(&self) -> Result<usize, String> {
+        match self.0.lock() {
```
notgull, line 60: I'm not sure if we want to lock an introduce contention here. Maybe we should use `try_lock` instead and return `0` if there is contention. Also, we shouldn't return `String` as that is a poor error type.

src/lib.rs
```diff
@@ -479,6 +479,40 @@ impl<T> Event<T> {
...
+    /// assert_eq!(event.total_listeners().unwrap(), 2);        
+    ///
+    /// drop(listener1);
+    /// drop(listener2);
+    /// assert_eq!(event.total_listeners().unwrap(), 0);        
+    ///
+    /// ```
+    #[cfg(feature = "std")]
+    #[inline]
+    pub fn total_listeners(&self) -> Result<usize, String> {
```
notgull, line 509: I would like to avoid using `String` as an error type. If we need to have an error I would introduce a `MutexPoisoned` newtype and use that as the error type. But other places in `event-listener` panic if the `Mutex` is poisoned, so we should continue that pattern.

notgull (changes requested):
I think the documentation should be modified with less guarantees as to the effectiveness of this function. For instance, the documentation should mention all of the following:

- This is just a snapshot of the number of listeners at this point in time. It is possible for the actual number to change at any point.
- The number should only ever be used as a hint.

src/std.rs
beckend: If we return 0, and the mutex is busy, and in matter fact there are listeners listening. The user then used a false number and executed logic based on that.

notgull: Alright, then that should be mentioned in the documentation.

beckend:
@notgull can you review it again?

src/lib.rs
```diff
@@ -479,6 +479,40 @@ impl<T> Event<T> {
 
         inner
     }
+
+    /// Return the listener count by acquiring a lock
```
notgull, line 483: Put a newline after this comment and add a period.

src/lib.rs
```diff
@@ -479,6 +479,40 @@ impl<T> Event<T> {
...
+    /// This is just a snapshot of the number of listeners at this point in time. It is possible for the actual number to change at any point.
```
notgull, line 484: Cap lines to 100 characters long.

src/std.rs
```diff
@@ -54,6 +54,14 @@ impl<T> List<T> {
             Err(_) => Err("<locked>"),
         }
     }
+
+    // Get the listener count by blocking
+    pub(crate) fn total_listeners_wait(&self) -> usize {
```
notgull, line 59: Name this function `total_listeners` instead.

notgull (changes requested):

src/std.rs
beckend: total_listeners is already just above.

notgull (approved):
Thanks!

Issue #118, opened 2024-02-07
Question about new API


mattklein123:
With the previous version I was able to do something like this: 

```
  async fn recv(&self) -> Vec<ParsedMetric> {
    let listener = EventListener::new();
    tokio::pin!(listener);

    loop {
      if let Some(metrics) = self.lifo.lock().pop_back() {
        return metrics;
      }

      if listener.is_listening() {
        listener.as_mut().await;
      } else {
        listener.as_mut().listen(&self.event);
      }
    }
  }
```

In the new version this seems to roughly translate to:

```
  async fn recv(&self) -> Vec<ParsedMetric> {
    loop {
      if let Some(metrics) = self.lifo.lock().pop_back() {
        return metrics;
      }

      listener!(self.event => listener);

      if let Some(metrics) = self.lifo.lock().pop_back() {
        return metrics;
      }

      listener.await
    }
  }
```

However this fails compile with:

```
rror: future cannot be sent between threads safely
   --> pulse-metrics/src/pipeline/processor/buffer/mod.rs:67:7
    |
67  |       tokio::spawn(async move {
    |       ^^^^^^^^^^^^ future created by async block is not `Send`
    |
    = help: within `event_listener::sys::Inner<()>`, the trait `std::marker::Send` is not implemented for `NonNull<event_listener::sys::Link<()>>`
note: future is not `Send` as this value is used across an await
   --> pulse-metrics/src/pipeline/processor/buffer/mod.rs:87:16
    |
81  |       listener!(self.event => listener);
    |       --------------------------------- has type `StackSlot<'_, ()>` which is not `Send`
...
87  |       listener.await
    |                ^^^^^ await occurs here, with `mut $listener` maybe used later
note: required by a bound in `tokio::spawn`
   --> /Users/mklein/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.36.0/src/task/spawn.rs:166:21
    |
164 |     pub fn spawn<F>(future: F) -> JoinHandle<F::Output>
    |            ----- required by a bound in this function
165 |     where
166 |         F: Future + Send + 'static,
    |                     ^^^^ required by this bound in `spawn`
```

I haven't dug into the guts but it seems like awaiting the future should somehow consume it such that this would not fail the Send check? Am I doing something wrong? Thank you.

notgull:
Nope, this is a bug, `StackSlot` should be `Send`.

mattklein123:
[...]

PR #119, opened 2024-02-08
StackSlot should be send

mattklein123:
Fixes https://github.com/smol-rs/event-listener/issues/118


mattklein123:
IMO it would be good to add a test that uses tokio and actually exercises this for real in that context. I can add one if you like.

notgull (approved):
Thanks!

> IMO it would be good to add a test that uses tokio and actually exercises this for real in that context. I can add one if you like.

Yes, that would be appreciated. Feel free to do it in another PR.

Issue #124, opened 2024-03-23
Upgrading v4 to v5 panic: Listener was never inserted into the list


jbr:
I'm unsure if I'm holding the new interface wrong somehow or if I've encountered a bug.

Given [this diff](https://github.com/trillium-rs/trillium/pull/595/files) that attempts to track the event-listener interface changes, I'm encountering a deterministic panic like:
```
event-listener-5.2.0/src/lib.rs:1251:36:
listener was never inserted into the list
```

and as far as I can tell the code reduces to

```rust
let n = AtomicUsize::from(1);
let event = Event::new();
let listener = event.listen();
loop {
    if 0 == n.load(Ordering::SeqCst) {
        return Poll::Ready(());
    };
    ready!(Pin::new(&mut self.listener).poll(cx));
}
```

Do I need to replace the listener every time I poll it? Am I holding this wrong? Is this a bug?

notgull:
It seems you are using the future like this:

```rust
let mut listener = event.listen();
loop {
    if 0 == n.load(Ordering::SeqCst) {
        return Poll::Ready(());
    };
    ready!(Pin::new(&mut listener).poll(cx));
} 
```

It should be used like this:

```rust
let mut listener = None;
loop {
    if 0 == n.load(Ordering::SeqCst) {
        return Poll::Ready(());
    };
    if let Some(listener) = listener.as_mut() {
        ready!(Pin::new(&mut listener).poll(cx));
        listener = None;
    } else {
        listener = Some(event.listen());
    }
}
```

Polling the `EventListener` after it has returned `Ready` will panic, and further polls should use a completely new `EventListener` instead.

The v4 API kept track of this state inside of the `EventListener,` but this [caused confusion](https://github.com/smol-rs/event-listener/issues/89) and was removed in v5.

notgull:
If you think this could be made clearer in documentation I would accept a PR.

jbr:
Would it make sense to just always return ready once it has ever been ready instead of panicking?

notgull:
Rust's documentation seems to imply that [a panic is the preferred option](https://docs.rs/futures/latest/futures/future/trait.Future.html#panics) in this case.

jbr:
[...]

PR #125, opened 2024-03-26
Make the panic message for a non-inserted listener clearer

notgull:
This commit makes the panic message for a listener that's not inserted into the linked list much clearer. The goal is to convey to the user that they may be `poll`ing the listener after it has completed.

This commit also fixes some new Clippy lints.

cc #124

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1248,7 +1248,7 @@ impl<T> RegisterResult<T> {
         match self {
             Self::Notified(tag) => Some(tag),
             Self::Registered => None,
-            Self::NeverInserted => panic!("listener was never inserted into the list"),
+            Self::NeverInserted => panic!("{}", NEVER_INSERTED_PANIC),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -68,6 +68,7 @@
 //! [`portable-atomic`]: https://crates.io/crates/portable-atomic
 
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::multiple_bound_locations)] // pin_project_lite can't work around this
```
zeenix: please try to keep separate logical changes in separate commits. It helps with reverting, when needed among other benefits of atomic commits.

src/lib.rs
```diff
@@ -1326,6 +1327,9 @@ impl TaskRef<'_> {
     }
 }
 
+const NEVER_INSERTED_PANIC: &str = "\
+EventListener was not inserted into the linked list, make sure you're not polling EventListener/listener! after it has finished";
```
zeenix, line 1330: nit: Please break this line to fit 100 chars.

zeenix (changes requested):
Just some nitpicks. :)

zeenix (approved):

PR #126, opened 2024-03-28
feat: Add a loom implementation for event-listener

### LINKED ISSUE #23: [Question] Integration with loom or shuttle?
Hi,

This package has been a lifesaver. I really unable to implement https://github.com/rmcgibbo/async-priority-channel (which whos API is basically copied completely from smol-rs/async-channel) without out.

One question: would it be straightforward/possible for event-listener to provide a hook to use [loom](https://github.com/tokio-rs/loom) or [shuttle](https://github.com/awslabs/shuttle). I'd like to be able to test my code which depends on event-listener using these techniques if possible. https://github.com/tokio-rs/loom/pull/220 seems to indicate that the latest release of loom now handles SeqCst which is probably necessary.

jbr:
The loom tests aren't particularly useful yet, but the crate compiles and runs under loom now.

Note: This crate does not currently compile under loom without std because concurrent-queue does not compile with that combination (can't find spin_loop).

closes #23

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -105,7 +105,10 @@ use {
 };
 
 use sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
-use sync::{Arc, WithMut};
+use sync::Arc;
+
+#[cfg(not(loom))]
+use sync::WithMut;
 
 use notify::{Internal, NotificationPrivate};
 pub use notify::{IntoNotification, Notification};
@@ -216,9 +219,16 @@ impl<T> Event<T> {
     ///
     /// let event = Event::<usize>::with_tag();
     /// ```
-    #[cfg(feature = "std")]
+    #[cfg(all(feature = "std", not(loom)))]
     #[inline]
     pub const fn with_tag() -> Self {
+        Self {
+            inner: AtomicPtr::new(ptr::null_mut()),
+        }
+    }
+    #[cfg(all(feature = "std", loom))]
+    #[inline]
+    pub fn with_tag() -> Self {
         Self {
             inner: AtomicPtr::new(ptr::null_mut()),
         }
@@ -543,7 +553,16 @@ impl Event<()> {
     /// let event = Event::new();
     /// ```
     #[inline]
+    #[cfg(not(loom))]
     pub const fn new() -> Self {
+        Self {
+            inner: AtomicPtr::new(ptr::null_mut()),
+        }
+    }
+
+    #[inline]
+    #[cfg(loom)]
+    pub fn new() -> Self {
         Self {
             inner: AtomicPtr::new(ptr::null_mut()),
         }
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -105,7 +105,12 @@ use {
 };
 
 use sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
-use sync::{Arc, WithMut};
+use sync::Arc;
+
+#[cfg(not(loom))]
+// WitMut is not needed under loom because loom::sync::atomic::AtomicPtr includes an intrinsic
+// with_mut that has the same interface as this crate's sync::WithMut::with_mut
```
notgull, line 112: This comment is not necessary; the `WithMut` trait was added entirely in preparation for `loom`.

src/lib.rs
```diff
@@ -216,13 +221,32 @@ impl<T> Event<T> {
...
+    /// Tagging cannot be implemented efficiently on `no_std`, so this is only available when the
+    /// `std` feature is enabled.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use event_listener::Event;
+    ///
+    /// let event = Event::<usize>::with_tag();
+    /// ```
```
notgull, line 242:
```suggestion

```

Docs aren't necessary here as `loom`-specific functions are not public API.

src/lib.rs
```diff
@@ -543,12 +567,30 @@ impl Event<()> {
...
 
+    /// Creates a new [`Event`].
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use event_listener::Event;
+    ///
+    /// let event = Event::new();
+    /// ```
```
notgull, line 585:
```suggestion

```

Ditto

notgull (changes requested):
I'm surprised `concurrent-queue` Loom tests are failing, but I guess that's due to `loom` not being checked in CI there.

notgull:
I added Loom to the CI in https://github.com/smol-rs/concurrent-queue/pull/62. It looks like it works. Maybe it's the optional feature flag that's gumming it up.

notgull:
I think `crossbeam` does this right. See here:

https://github.com/crossbeam-rs/crossbeam/blob/9e8596105bc9a6b343918b6ad1c9656dc24dc4f9/crossbeam-epoch/Cargo.toml#L30-L34

and here:

https://github.com/crossbeam-rs/crossbeam/blob/9e8596105bc9a6b343918b6ad1c9656dc24dc4f9/crossbeam-epoch/Cargo.toml#L39-L44

jbr:
> I'm surprised `concurrent-queue` Loom tests are failing, but I guess that's due to `loom` not being checked in CI there.

To be clear, it's _not_ that the concurrent-queue loom tests are failing, or even that concurrent-queue can't build under loom. This PR for event listener correctly builds concurrent-queue under loom and runs tests against concurrent-queue loom.

The [issue](https://github.com/smol-rs/concurrent-queue/issues/64) is this specific scenario for concurrent-queue, which may or may not even matter: `RUSTFLAGS="--cfg=loom" cargo build --no-default-features --features loom` (in concurrent-queue)

notgull (approved):
Thanks!

Issue #123, opened 2024-03-03
Miri reports a deadlock in `async-lock`


notgull:
https://github.com/smol-rs/async-lock/pull/80#issuecomment-1975145930

Given that this only cropped up after the v5.0 update, it's probably this crate's fault.

notgull:
Verified that this bug was introduced in commit smol-rs/async-lock@ff4c96fe4dc72f423 or commit 86b7780, as the combination of these two commits are the first commits that observe this bug.

Diff applies to `async-lock`:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 88a24be..8062767 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ categories = ["asynchronous", "concurrency"]
 exclude = ["/.*"]
 
 [dependencies]
-event-listener = { version = "5.0.0", default-features = false }
+event-listener = { version = "4.0.0", default-features = false }
 event-listener-strategy = { version = "0.5.0", default-features = false }
 pin-project-lite = "0.2.11"
 
@@ -31,3 +31,7 @@ waker-fn = "1.1.0"
 
 [target.'cfg(target_family = "wasm")'.dev-dependencies]
 wasm-bindgen-test = "0.3"
+
+[patch.crates-io]
+event-listener = { git = "https://github.com/smol-rs/event-listener", rev = "86b7780" }
+event-listener-strategy = { path = "../event-listener-strategy" }
diff --git a/src/lib.rs b/src/lib.rs
index 6d64aa3..4debf8c 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,13 +73,13 @@ macro_rules! pin {
 
 mod barrier;
 mod mutex;
-mod once_cell;
+//mod once_cell;
 mod rwlock;
 mod semaphore;
 
 pub use barrier::{Barrier, BarrierWaitResult};
 pub use mutex::{Mutex, MutexGuard, MutexGuardArc};
-pub use once_cell::OnceCell;
+//pub use once_cell::OnceCell;
 pub use rwlock::{
     RwLock, RwLockReadGuard, RwLockReadGuardArc, RwLockUpgradableReadGuard,
     RwLockUpgradableReadGuardArc, RwLockWriteGuard, RwLockWriteGuardArc,
```

Diff applied to `event-listener-strategy`:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index fe6bf0d..13e6170 100644
[...]

PR #127, opened 2024-03-31
Use a fresh parker on Miri

notgull:
Miri appears to have issues tracking the unparker when it's stored in a
thread-local variable. For now, we can just create a fresh parker for
every wait when it comes to Miri.

cc #123

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1101,19 +1101,30 @@ impl<T, B: Borrow<Inner<T>> + Unpin> InnerListener<T, B> {
             static PARKER: (Parker, Task) = parker_and_task();
         }
 
-        // Try to borrow the thread-local parker/unparker pair.
-        PARKER
-            .try_with({
-                let this = self.as_mut();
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1101,19 +1101,30 @@ impl<T, B: Borrow<Inner<T>> + Unpin> InnerListener<T, B> {
...
+                    let this = self.as_mut();
+                    |(parker, unparker)| this.wait_with_parker(deadline, parker, unparker.as_task_ref())
+                })
+                .unwrap_or_else(|_| {
+                    // If the pair isn't accessible, we may be being called in a destructor.
+                    // Just create a new pair.
+                    self.as_mut().wait_with_fresh_parker(deadline)
+                })
+        } else {
+            // Miri seems to have issues keeping track of thread-local wakers. Just use a fresh pair.
```
taiki-e, line 1117: The comment says "Miri ...", but this is `else` block of `if cfg!(miri)` (i.e., this branch is only invoked in non-Miri cases).

notgull:
On further analysis I don't think this is the issue

PR #130, opened 2024-04-11
Roll algorithm back to v2.5.3, then add back features to avoid a breaking change

### LINKED ISSUE #109: Remove concurrent-queue dependency
`concurrent-queue` is a pretty heavy dependency for a crate as small as `event-listener`. It would be nice if we were able to avoid using it.

notgull:
I have to apologize. I know it sounds self-deprecating, but really, I've been over the line on this particular issue. This should hopefully fix it for now.

The new implementation for the event listener, #51, has been plagued with bugs from its inception. Frankly, I've gotten sick of papering over these bugs, and I know that prod customers are almost certainly impacted but haven't reported it.

This PR reverts the implementation of `event-listener` back to version 2.5.3; before I made any of the sweeping architectural changes to the crate. Then I re-add new changes to this crate in a "path of least resistance" fashion. The goal is to change the internal algorithm as little as possible, as we know this algorithm is stable and robust.

The only real diversion from this path is my adding of a `no_std` implementation. I've added a `no_std` implementation based on an atomic linked list built on an amortized atomically expanded vector. Since it doesn't use `concurrent-queue` it fixes #109.

Let me run a "crater test" on the `smol` ecosystem before this is merged. Expect a "post mortem" at some point in the near future.


notgull:
Paging @smol-rs/admins for reviewing this PR. Also paging @mamaicode, @jbr and @beckend as I've re-written some of your commits here.

benches/bench.rs
```diff
@@ -8,14 +6,16 @@ const COUNT: usize = 8000;
 fn bench_events(c: &mut Criterion) {
     c.bench_function("notify_and_wait", |b| {
         let ev = Event::new();
-        let mut handles = Vec::with_capacity(COUNT);
-
         b.iter(|| {
-            handles.extend(iter::repeat_with(|| ev.listen()).take(COUNT));
+            let mut handles = Vec::with_capacity(COUNT);
```
fogti, line 10: I'd prefer it if the iterator filling would be kept (and just `.collect()`ed)

src/lib.rs
```diff
@@ -67,74 +68,159 @@
...
+///     }
+///
+///     // Wait for a notification and continue the loop.
+///     listener.wait();
+/// }
+/// ```
+#[macro_export]
+macro_rules! listener {
+    ($event:expr => $listener:ident) => {
+        // TODO: Stack allocation.
```
fogti, line 221: oh

src/lib.rs
```diff
@@ -851,701 +734,171 @@ pub trait Listener<T = ()>: Future<Output = T> + __sealed::Sealed {
...
+    ) -> R {
+        f(ptr.get_mut())
     }
+}
 
-    impl<T> WithMut for atomic::AtomicPtr<T> {
-        type Output = *mut T;
+#[cfg(loom)]
+mod loom {
+    pub(crate) use loom::cell;
```
fogti, line 874: hmm how does symbol lookup / precedence work here?

src/notify.rs
```diff
@@ -620,3 +613,13 @@ mod __private {
...
     impl<N: super::NotificationPrivate + ?Sized> Sealed for N {}
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    #[should_panic]
+    fn negative_notification_should_panic() {
+        let event = crate::Event::new();
+        event.notify(-1);
```
fogti, line 623: why were negative notifications allowed at all (and not prevented by using a type that doesn't have these)?

tests/counter.rs
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    /// Wait for the counter to be incremented.
+    async fn change(&self) -> usize {
+        let original = self.counter.load(Ordering::Acquire);
+        let mut current = original;
+
+        loop {
+            if current != original {
```
fogti, line 31: maybe move this below the "Try again." part

src/notify.rs
notgull: Because by default numeric literals passed into `notify` will resolve to `i32`, so to handle this common case we need to handle this number.

src/lib.rs
notgull: Yeah stack allocation isn't really possible with the current API. Maybe at some point later...

zeenix:
While I understand why you did it the way you did (to be able to better test/bisect the changes) but it's very hard to review a big bunch of commits like this. Also since you'll likely be squashing, can you already squash all the reverting commits into one (you can add in the description if you only reverted specific commits or it's a hard reset) and then add the new features on top (with each commit describing the change and why it's different now). After that, please do not squash merge when it's time to merge, to maintain the history.

I'm sorry for asking you to do all that but I've very limited time these days for smol (as you might have noticed). :(

notgull:
> While I understand why you did it the way you did (to be able to better test/bisect the changes) but it's very hard to review a big bunch of commits like this. Also since you'll likely be squashing, can you already squash all the reverting commits into one (you can add in the description if you only reverted specific commits or it's a hard reset) 

The reverts are all squashed into https://github.com/smol-rs/event-listener/pull/130/commits/636804689805447b094a0eecf40c165228815814. The rest of the commits are re-adding the functionality.

> then add the new features on top (with each commit describing the change and why it's different now).

Done.

> After that, please do not squash merge when it's time to merge, to maintain the history.

Sounds good!

> I'm sorry for asking you to do all that but I've very limited time these days for smol (as you might have noticed). :(

All good, no worries

tests/counter.rs
notgull: The event listener pattern generally assumes that the condition has been already checked before the listener is inserted.

src/lib.rs
notgull: Only one set of symbols are loaded based on which features are enabled

.github/workflows/ci.yml
```diff
@@ -27,80 +26,42 @@ defaults:
...
+        if: startsWith(matrix.rust, 'nightly')
+        run: cargo check -Z features=dev_dep
+      - run: cargo test
 
   msrv:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # When updating this, the reminder to update the minimum supported
+        # Rust version in Cargo.toml.
```
taiki-e, line 50: I don't feel many of the changes to CI need to be reverted.

beckend:
@notgull what stops you from merging this?

notgull:
> @notgull what stops you from merging this?

I'm waiting for reviews from other `smol-rs` members. I'm aware it's a big PR and that they don't have much time to review it.

notgull:
Actually, come to think of it this PR might be a little much.

@smol-rs/admins Would it be preferred if I split this up into smaller PRs, starting with an initial rollback followed by adding on more features?

zeenix:
> @smol-rs/admins Would it be preferred if I split this up into smaller PRs, starting with an initial rollback followed by adding on more features?

I can't say for others but what makes this PR hard to review for me, is it's hard to see what really changed in the end. I understand why you did it the way you did, to be able to figure out where the regression(s) happened but now that you know, I think it would be much better if we don't go back in history, but instead add the needed fixes on top of the current main branch (i-e not doing rollbacks, revert of specific commits is of course good). WDYT?

notgull:
> I understand why you did it the way you did, to be able to figure out where the regression(s) happened but now that you know, I think it would be much better if we don't go back in history, but instead add the needed fixes on top of the current main branch (i-e not doing rollbacks, revert of specific commits is of course good). WDYT?

The issue is that I'm not 100% sure what the current regressions are caused by, and the current advantages of the current implementation are not worth the effort of my part to investigate them.

beckend:
Of you have proven stability by adding all these tests, why not call it good enough of a change and merge it.

zeenix:
> The issue is that I'm not 100% sure what the current regressions are caused by, and the current advantages of the current implementation are not worth the effort of my part to investigate them.

Fair enough but then how do we ensure that the regressions weren't re-added by the "new" commits? For all we know, there could be a timing issue and tests don't catch the issue(s) still present. Finding out the issue and fixing it, would be the only certain way of actually fixing the problems. I would hate for us to have to break API again and again.

notgull:
> Fair enough but then how do we ensure that the regressions weren't re-added by the "new" commits? For all we know, there could be a timing issue and tests don't catch the issue(s) still present.

I'm using the `async-lock` rwlock contention tests as a benchmark here. This current code passes this test and `master` does not. I'll admit that it's not a perfect benchmark but I think it generally works as an indicator.

But, fair enough. I'll do a deep dive into the current code and see if I can figure out what the issue is.

zeenix:
> But, fair enough. I'll do a deep dive into the current code and see if I can figure out what the issue is.

I appreciate your very constructive attitude to my rather blunt feedback. :+1:

notgull:
cc https://github.com/smol-rs/event-listener/pull/133

notgull:
Obsoleted by #139

PR #134, opened 2024-05-19
Make this crate work on embedded targets

BlueZeeKing:
- Document `std` feature
- Set `concurrent-queue/portable-atomic` feature correctly

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,13 +62,13 @@
 //!
 //! # Features
 //!
+//! - The `std` feature (enabled by default) enables the use of the Rust standard library. Disable it for `no_std`
+//! support
+//!
 //! - The `portable-atomic` feature enables the use of the [`portable-atomic`] crate to provide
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -66,6 +66,9 @@
 //!   atomic operations on platforms that don't support them.
 //!
 //! [`portable-atomic`]: https://crates.io/crates/portable-atomic
+//!
+//! - The `std` feature enables the use of the Rust standard library. Disable it for `no_std`
```
taiki-e, line 70:
```suggestion
//! - The `std` feature (enabled by default) enables the use of the Rust standard library. Disable it for `no_std`
```

Also, it might be better to swap the order of this and `portable-atomic` feature to match the order in Cargo.toml.

taiki-e:
Thanks! Rebasing should fix the CI failure.

taiki-e (approved):

PR #135, opened 2024-05-19
Fix new nightly Clippy warnings

notgull:
- Suppress warning from the cfg(loom) directive by having it emit
  "rustc-check-cfg" in build.rs
- Remove usage of legacy "std::usize" module.
- Ignore "clippy::multiple_bound_locations". We use multiple bounds as a
  workaround for an issue in pin-project-lite where only one bound can be
  assigned per location. The phrasing of pin-project-lite's README implies that
  this issue is WONTFIX. To quote:

> This library does not tackle as expansive of a range of use cases as
> pin-project does. If your use case is not already covered, please use
> pin-project.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,7 @@
+//! Emits the necessary "checked" compile time flag warning for "loom".
+//!
+//! Without this, the compiler complains about the unspecified "loom" flag.
+
+fn main() {
+    println!("cargo:rustc-check-cfg=cfg(loom)");
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,6 +68,7 @@
 //! [`portable-atomic`]: https://crates.io/crates/portable-atomic
 
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::multiple_bound_locations)] // This is a WONTFIX issue with pin-project-lite
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
 #![doc(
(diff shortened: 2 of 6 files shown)


notgull:
MIRI failure is tracked in https://github.com/smol-rs/event-listener/pull/133

build.rs
```diff
@@ -0,0 +1,7 @@
+//! Emits the necessary "checked" compile time flag warning for "loom".
+//!
+//! Without this, the compiler complains about the unspecified "loom" flag.
+
+fn main() {
+    println!("cargo:rustc-check-cfg=cfg(loom)");
```
taiki-e, line 6: I would like to use https://github.com/rust-lang/cargo/pull/13913 's config option instead of adding build script.

src/lib.rs
```diff
@@ -68,6 +68,7 @@
 //! [`portable-atomic`]: https://crates.io/crates/portable-atomic
 
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::multiple_bound_locations)] // This is a WONTFIX issue with pin-project-lite
```
taiki-e, line 71: If this lint is triggered for code generated by external macros, it is definitely clippy bug.

taiki-e: Ok, I found https://github.com/rust-lang/rust-clippy/issues/12552.

taiki-e: Oh, looking at your PR description, this lint may also need to be allowed on the pin-project-lite side, but it's not clear which one is actually causing a warning here.

build.rs
notgull: Tracked in #136

PR #136, opened 2024-05-19
Use Cargo.toml lint fix instead of build.rs

notgull:
https://github.com/smol-rs/event-listener/pull/135#discussion_r1606080083


taiki-e:
Thanks! Could you remove the build script?

PR #139, opened 2024-05-27
Disable early-out notify optimizations

### LINKED ISSUE #138: During racy initialization, a race can happen where a notification is dropped
The following logic exists in `notify`:

https://github.com/smol-rs/event-listener/blob/fdbe437699680bb5e402bc89aa2a0dd12a1ebc25/src/lib.rs#L440

The intent is to make it so `notify`ing an `Event` that has never has an `EventListener` is a single atomic operation and nothing else. However, a race condition can happen here where a notification is dropped even when there is a listener available.

- Thread 0 calls `listen()`. As part of `listen()` the `inner()` function is called, which allocates the `Arc<Inner>` and `compare_exchanges` it into the `inner` pointer.
- However, between the `load` of the `inner` pointer and the final `compare_exchange`, Thread 0 is pre-empted and Thread 1 starts.
- Thread 1 calls `notify`, which then calls `try_inner`, which `load`s the `inner` pointer. Since it hasn't been updated yet this returns `0`, which the `notify` function interprets as the event not listening yet.
- Thread 1 finished `notify` by notifying no listeners.

If Thread 0 pre-empted Thread 1 while Thread 1 was updating some atomic state, this can lead to a deadlock.

notgull:
The notify() function contains two optimizations:

- If the `inner` field is not yet initialized (i.e. no listeners have been
  created yet), it returns `0` from the function early.
- If the atomic `notified` field indicates that the current notification would
  do nothing, it returns `0` early as well.

`loom` testing has exposed race conditions in these optimizations that can cause
notifications to be missed, leading to deadlocks. This commit removes these
optimizations in the hope of preventing these deadlocks. Ideally we can restore
them in the future.

Closes #138
cc #130 and #133

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -111,7 +111,7 @@ use sync::Arc;
 #[cfg(not(loom))]
 use sync::WithMut;
 
-use notify::{Internal, NotificationPrivate};
+use notify::NotificationPrivate;
 pub use notify::{IntoNotification, Notification};
 
 /// Inner state of [`Event`].
@@ -437,21 +437,8 @@ impl<T> Event<T> {
         // Make sure the notification comes after whatever triggered it.
         notify.fence(notify::Internal::new());
 
-        if let Some(inner) = self.try_inner() {
-            let limit = if notify.is_additional(Internal::new()) {
-                usize::MAX
-            } else {
-                notify.count(Internal::new())
-            };
-
-            // Notify if there is at least one unnotified listener and the number of notified
-            // listeners is less than `limit`.
-            if inner.needs_notification(limit) {
-                return inner.notify(notify);
-            }
-        }
-
-        0
+        let inner = unsafe { &*self.inner() };
+        inner.notify(notify)
     }
 
     /// Return a reference to the inner state if it has been initialized.
--- a/src/no_std.rs
+++ b/src/no_std.rs
@@ -45,11 +45,6 @@ impl<T> crate::Inner<T> {
         drop(self.try_lock());
     }
 
-    pub(crate) fn needs_notification(&self, _limit: usize) -> bool {
-        // TODO: Figure out a stable way to do this optimization.
-        true
-    }
-
     /// Add a new listener to the list.
     ///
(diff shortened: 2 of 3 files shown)


notgull:
I've tested `async-lock` with a patched version of `event-listener` with this PR and the MIRI tests pass, even with the previously ignored tests un-ignored. In addition I'm relatively sure it passes Loom tests (https://github.com/smol-rs/async-lock/pull/86). However I can't be 100% sure; my max-preemptions count is set very low as I do not have access to hardware that can conduct full loom tests at the moment.

I'm unclear why this bug is manifesting now; pretty much all versions of `event-listener` contain this code and v5 didn't really touch it. Maybe it just offset the timing *just right* for the race condition to start showing up in normal code.

zeenix (commented):
Great work figuring this one! 👍 Wouldn't removing these optimizations mean that we'll wake up tasks unnecessarily? 🤔

notgull:
No, these optimizations were there to add fast paths that didn't lock the inner mutex and traverse the linked list. This operation is somewhat expensive so this optimization avoided doing that if it could. Removing it won't wake up any additional tasks.

fogti (approved):

zeenix (approved):
Thanks for explaining. LGTM then.

src/lib.rs
```diff
@@ -437,21 +437,8 @@ impl<T> Event<T> {
...
-
-            // Notify if there is at least one unnotified listener and the number of notified
-            // listeners is less than `limit`.
-            if inner.needs_notification(limit) {
-                return inner.notify(notify);
-            }
-        }
-
-        0
+        let inner = unsafe { &*self.inner() };
```
taiki-e, line 440: I guess try_inner -> inner change causes UB (null-ptr-deref) on `Event::new().notify()`.

notgull: The `inner` function initializes the inner pointer if it's null, so this shouldn't be an issue.

taiki-e: Oh, sorry, you're right.

PR #140, opened 2024-05-29
v5.3.1

notgull:
- Disable some optimizations that, in rare conditions, can cause race conditions
  causing notifications to be dropped. (#139)
- Ensure the portable-atomic feature is set properly. (#134)
- Update `portable-atomic-util` to v0.2.0. (#132)
- Document the std feature. (#134)


taiki-e:
Thanks! We may want to add https://github.com/smol-rs/event-listener/pull/132 to changelog too.

Otherwise LGTM.

taiki-e (approved):

PR #144, opened 2024-10-18
Use "v2.0.0" branch for security check

notgull:


fogti:
WASM is broken, huh?

notgull:
Yeah I think that's recent.

fogti (approved):

PR #145, opened 2024-10-29
Fix missed notification in Mutex example

### LINKED ISSUE #143: This seems actually not stable(Dead lock)
https://github.com/beckend/event-listener/tree/event_listener_bug

`cargo run --release --example mutex`

https://github.com/beckend/event-listener/blob/event_listener_bug/examples/mutex.rs#L127-L151

It dead locks.

notgull:
I forgot to run notify() when I rewrote this example without unsafe
code. It looks like it prevents deadlocks in this mutex implementation.

I believe this fixes #143. @beckend Can you confirm?

### DIFF
--- a/examples/mutex.rs
+++ b/examples/mutex.rs
@@ -37,7 +37,10 @@ mod example {
 
         /// Attempts to acquire a lock.
         fn try_lock(&self) -> Option<MutexGuard<'_, T>> {
-            self.data.try_lock().map(MutexGuard)
+            self.data.try_lock().map(|l| MutexGuard {
+                lock_ops: &self.lock_ops,
+                locked: Some(l),
+            })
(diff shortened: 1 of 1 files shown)


examples/mutex.rs
```diff
@@ -107,19 +110,29 @@ mod example {
     }
 
     /// A guard holding a lock.
-    struct MutexGuard<'a, T>(Locked<'a, T>);
+    struct MutexGuard<'a, T> {
+        mutex: &'a Mutex<T>,
```
fogti, line 114: Given that we only access `mutex.lock_ops`, wouldn't it be more appropriate to just place a pointer to `lock_ops` here?

fogti (approved):

PR #148, opened 2024-11-17
Add an efficient implementation based on critical-section

notgull:
The current no-std implementation is somewhat ineffecient, potentially
broken and relies on some slow data structures. The main reason why we
use this implementation is because we don't have a way to "lock" the
linked list, since there's no way to lock things in no-std. However,
many embedded platforms have a concept of a "critical section" that can
be used to exclusively lock something.

This PR adds an option to the "std" implementation that replaces the
existing Mutex with a usage of the critical-section crate. This is
enabled with the "critical-section" feature. This allows us to have the
advantages of the std-based implementation without needing to rely on
std for platforms that don't have it.


src/std.rs
```diff
@@ -1,20 +1,32 @@
...
+//! is secured using either a libstd mutex or a critical section.
 
 use crate::notify::{GenericNotify, Internal, Notification};
 use crate::sync::atomic::Ordering;
 use crate::sync::cell::{Cell, UnsafeCell};
-use crate::sync::{Mutex, MutexGuard};
 use crate::{RegisterResult, State, TaskRef};
 
+#[cfg(feature = "critical-section")]
+use core::cell::RefCell;
```
zeenix, line 12: nitpick: I have learnt that it's better to not import conditionally as it then increases the `cfg` lines. E.g I personally favour

```rust
#[cfg(features = "xyz")]
xyz::Xyz::foo();

...

#[cfg(features = "xyz")]
xyz::Xyz::bar();
```

over

```rust
#[cfg(features = "xyz")]
use xyz::Xyz;

...

#[cfg(features = "xyz")]
Xyz::blah();

...

#[cfg(features = "xyz")]
Xyz::bar();
```

Of course it's a personal taste so up to you. :)

src/lib.rs
```diff
@@ -85,8 +85,11 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std as alloc;
 
-#[cfg_attr(feature = "std", path = "std.rs")]
-#[cfg_attr(not(feature = "std"), path = "no_std.rs")]
+#[cfg_attr(any(feature = "std", feature = "critical-section"), path = "std.rs")]
+#[cfg_attr(
+    not(any(feature = "std", feature = "critical-section")),
+    path = "no_std.rs"
+)]
```
zeenix, line 92: hmm.. I'm a bit confused. I thought `critical-section` is for no_std but here it seems we're `std` module for both? :thinking:

zeenix (commented):
Cool stuff! I had planned to look into critical-sections for async-broadcast but never got around to it. However, since a-b depends on event-listener, this is required anyway. I just have a confusion and a nitpick to offer. :)

src/lib.rs
notgull: Maybe I should rename the `std` and `no_std` modules to something else, now that they don't correspond to `std` and `no_std` more. Maybe `intrusive` vs `slab`, since that's what the implementation is based on.

zeenix: > Maybe `intrusive` vs `slab`, since that's what the implementation is based on.

Sounds good to me.

src/std.rs
notgull: I prefer to avoid using full paths in the body of the code.

zeenix: normally, so do I. It's just that the code becomes much harder to follow if imports are conditional. Anyway, as I wrote, up to you. :)

src/intrusive.rs
```diff
@@ -36,27 +48,77 @@ struct Inner<T> {
...
+    }
+
     /// Get the total number of listeners with blocking.
+    #[cfg(all(feature = "std", not(feature = "critical-section")))]
     pub(crate) fn total_listeners(&self) -> usize {
         self.0.lock().unwrap_or_else(|e| e.into_inner()).len
     }
+
+    /// Get the total number of listeners with blocking.
+    #[cfg(all(feature = "std", feature = "critical-section"))]
```
fogti, line 84: This line looks wrong.
Also, I think it would be a good idea to define `total_listeners` explicitly, and `try_total_listeners` via `total_listeners` in the `critical-section` case.

src/intrusive.rs
```diff
@@ -274,11 +337,13 @@ impl<T> Inner<T> {
     }
 }
 
+#[cfg(all(feature = "std", not(feature = "critical-section")))]
```
fogti, line 340: imo it would be a good idea to put `ListLock` into a separate file and mark the whole file with `#[cfg(...)]`, this would make this easier to read.

src/lib.rs
```diff
@@ -1359,7 +1365,7 @@ mod sync {
     #[cfg(feature = "portable-atomic")]
     pub(super) use portable_atomic_util::Arc;
 
-    #[cfg(all(feature = "std", not(loom)))]
+    #[cfg(all(feature = "std", not(feature = "critical-section"), not(loom)))]
```
fogti, line 1368: why wasn't the `critical_section::Mutex` import put here, too?

fogti (changes requested):
Great work, there is at least one small problem, tho.

src/intrusive.rs
notgull: Good point, fixed.

src/intrusive.rs
notgull: I've inlined the entire construct into `with_inner`.

src/lib.rs
notgull: Because `critical-section` doesn't rely on `loom` intrinsically.

fogti: hm `std` also doesn't rely on `loom` intrinsically; or how is that meant?/ what does that mean?

I just thought it should belong here because that part does accumulate definitions which are basically "alternatives" between different features used throughout the code, and the underlying `Mutex` impl imo is one of those "switchable" things (unfortunately not with a 1-to-1 common/shared sufficient API)

src/intrusive.rs
```diff
@@ -57,9 +56,13 @@ impl<T> List<T> {
         };
 
         #[cfg(feature = "critical-section")]
-        let inner = RefCell::new(inner);
+        {
+            let inner = RefCell::new(inner);
+            Self(critical_section::Mutex::new(inner))
```
fogti, line 61: why is `inner` not inlined?

fogti (approved):

src/intrusive.rs
fogti: oh that's quite neat, thank you.

src/lib.rs
notgull: With std you need to replace std's primitives with loom, with critical-section you can just use a CS based on Loom.

src/intrusive.rs
notgull: Because im silly, will fix

src/intrusive.rs
```diff
@@ -1,20 +1,31 @@
...
-use core::ops::{Deref, DerefMut};
 use core::pin::Pin;
 use core::ptr::NonNull;
 
-pub(super) struct List<T>(Mutex<Inner<T>>);
+pub(super) struct List<T>(
+    /// libstd-based implementation uses a normal Muetx to secure the data.
+    #[cfg(all(feature = "std", not(feature = "critical-section")))]
+    crate::sync::Mutex<Inner<T>>,
+    /// CS-based implementation uses a CS cell that wraps a RefCell.
```
fogti, line 25: I think this should spell out `critical section` instead of `CS` at least once, otherwise it might lead to confusion.

src/intrusive.rs
```diff
@@ -36,67 +47,141 @@ struct Inner<T> {
...
             head: None,
             tail: None,
             next: None,
             len: 0,
             notified: 0,
-        }))
+        };
+
+        #[cfg(feature = "critical-section")]
+        {
```
fogti, line 59: unnecessary parentheses

notgull: You'd think, but when I remove these braces it throws a compile error.

fogti (approved):

PR #149, opened 2024-12-25
feat: Add scheduling options for listener lists.

jevolk:
Two schedulers are available:
- FIFO: The original round-robin queuing; listeners are inserted at the back.
- LIFO: The new most-recent queuing; listeners are inserted at the front.

LIFO queuing is beneficial for cache-efficiency with workloads that are tolerant of starvation. The same listener is repeatedly drawn from the list until the load dictates additional listeners be drawn from the list. These listeners expand outward as a "hot set" for optimal reuse of resources rather than continuously drawing from the coldest resources in a FIFO schedule.

### DIFF
--- a/src/intrusive.rs
+++ b/src/intrusive.rs
@@ -6,7 +6,7 @@
 use crate::notify::{GenericNotify, Internal, Notification};
 use crate::sync::atomic::Ordering;
 use crate::sync::cell::{Cell, UnsafeCell};
-use crate::{RegisterResult, State, TaskRef};
+use crate::{QueueStrategy, RegisterResult, State, TaskRef};
 
 #[cfg(feature = "critical-section")]
 use core::cell::RefCell;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -127,6 +127,16 @@ use sync::WithMut;
 use notify::NotificationPrivate;
 pub use notify::{IntoNotification, Notification};
 
+/// Queue schedule of a listener.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Sched {
```
notgull, line 132: This should be `#[non_exhaustive]`.

src/lib.rs
```diff
@@ -127,6 +127,16 @@ use sync::WithMut;
 use notify::NotificationPrivate;
 pub use notify::{IntoNotification, Notification};
 
+/// Queue schedule of a listener.
+#[derive(Clone, Copy, Debug, PartialEq)]
```
notgull, line 131: I would prefer a name more descriptive than `Sched`. Maybe `QueueStrategy`?

src/intrusive.rs
```diff
@@ -248,6 +222,51 @@ impl<T> crate::Inner<T> {
...
+                None => *replacing = Some(entry.into()),
+                Some(t) if self.sched == Sched::Lifo => unsafe {
+                    t.as_ref().prev.set(Some(entry.into()))
+                },
+                Some(t) if self.sched == Sched::Fifo => unsafe {
+                    t.as_ref().next.set(Some(entry.into()))
+                },
+                Some(_) => unimplemented!("unimplemented scheduling type"),
+            };
+        }
```
notgull, line 257: Would prefer it if this block became:

```rust
match (mem::replace(replacing, Some(entry.into())), self.sched) {
    /* ... */
}
```

notgull (changes requested):

PR #154, opened 2025-06-26
feat: Add queuing strategies for listener lists.

jevolk:
Two strategies are available:
- FIFO: The original round-robin queuing; listeners are inserted at the back.
- LIFO: The new most-recent queuing; listeners are inserted at the front.

LIFO queuing is beneficial for cache-efficiency with workloads that are tolerant of starvation. The same listener is repeatedly drawn from the list until the load dictates additional listeners be drawn from the list. These listeners expand outward as a "hot set" for optimal reuse of resources rather than continuously drawing from the coldest resources in a FIFO schedule.

(Previously #149 but closed automatically when repository moved)

### DIFF
--- a/src/intrusive.rs
+++ b/src/intrusive.rs
@@ -6,7 +6,7 @@
 use crate::notify::{GenericNotify, Internal, Notification};
 use crate::sync::atomic::Ordering;
 use crate::sync::cell::{Cell, UnsafeCell};
-use crate::{RegisterResult, State, TaskRef};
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks for the PR. I haven't checked the implementation details, but in any case, there is no possibility of accepting a PR without any testing.

notgull (approved):
Looks good to me! Please rebase on latest `master`.

notgull:
Looks like CI is failing now...

jevolk:
Yea, do you need lifo implemented for slab?

